PR #2, opened 2020-04-08
Allow external DB instead of always installing etcd

JoelColledge:
This requires LINSTOR v1.6.0+.

### Usage

Set up a postgres DB with service name "postgres" and the following configuration:

```
  POSTGRES_DB: postgresdb
  POSTGRES_USER: postgresadmin
  POSTGRES_PASSWORD: admin123
```

Add this to the helm install command:

```
--set "operator.controllerSet.spec.dbConnectionURL=jdbc:postgresql://postgres/postgresdb?user=postgresadmin&password=admin123" --set etcd.enabled=false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,27 @@ Lastly, edit the storage nodes' LVM VG and LV names in examples/piraeus-operator
 kubectl create -f examples/piraeus-operator-part-2.yaml
 ```
 
+## Upgrading
+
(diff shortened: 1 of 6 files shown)


rck:
hm, having it still named `etcdUrl` does not feel right to me. Better break the API now, when it is easy and there are not too many users. But you are the maintainer, so your decision.

This should be documented in the `README.md` (basically copy/paste from the PR message).

PR #7, opened 2020-05-05
Allow helm to be more flexible when creating storage pools

WanzenBug:
Directly pass storage pool configuration from helm to the operator. The operator already supports some important features:
* multiple storage pools of the same type
* no automatic storage pool creation

With this change, the configuration given to helm is basically directly copied to the NodeSet definition.


WanzenBug:
Thinking about this again, its probably better to just "paste" the whole `storagePools` block directly from the helm values. There is no validation of the helm values right now anyway...

README.md
```diff
@@ -56,14 +56,32 @@ The operator can be deployed with Helm v3 chart in /charts as follows:
...
 
-- Configure the LVM VG and LV names by setting the following values in a local
-  values file or by using `--set` with the `helm install` command below.
-
-    ```
-    operator.nodeSet.spec.lvmPoolVgName=drbdpool # <- Volume Group name of the thick storage pool
-    operator.nodeSet.spec.lvmThinPoolVgName=drbdpool # <- Volume Group name of the thin storage pool
-    operator.nodeSet.spec.lvmThinPoolLvName=thinpool # <- Logical Volume name of the thin pool
-    ```
+- Configure storage pools. By default, helm creates two pools (see [values.yaml]):
```
JoelColledge, line 59: I actually think we should change the default to not create any pools. It would be unusual that the user wants to have precisely the 2 pools defined in `values.yaml` and we're still in 0.x version range so we don't need to worry too much about making breaking changes.

README.md
```diff
@@ -56,14 +56,32 @@ The operator can be deployed with Helm v3 chart in /charts as follows:
...
+  ```yaml
```
JoelColledge, line 60: Maybe we should put these instructions and this example in their own subsection to avoid the list getting too long and hard to follow.

JoelColledge:
This is a very clean and powerful solution :+1:

README.md
WanzenBug: Updated the default values for helm. During that I also noticed that there was a second copy of the CRDs hidden in the "non-helm" deployment examples. I removed them from the examples and added instructions to install them by hand.

README.md
WanzenBug: Moved

README.md
```diff
@@ -108,6 +102,64 @@ etcd cluster by adding the following to the Helm install command:
...
+
+There are two ways to configure storage pools
+
+#### On install
+On install time, by setting the value of `operator.nodeSet.spec.storagePools` when running helm install.
+
+First create a file with the storage configuration like:
+```yaml
+operator:
+nodeSet:
```
JoelColledge, line 134: One level of indentation appears to have gone missing here.

README.md
```diff
@@ -108,6 +102,64 @@ etcd cluster by adding the following to the Helm install command:
...
+    - name: lvm-thin
+      thinVolume: thinpool
+      volumeGroup: drbdpool
+
+```
+
+There are two ways to configure storage pools
+
+#### On install
+On install time, by setting the value of `operator.nodeSet.spec.storagePools` when running helm install.
```
JoelColledge, line 129: *At* install time

README.md
```diff
@@ -108,6 +102,64 @@ etcd cluster by adding the following to the Helm install command:
 --set etcd.enabled=false --set "operator.controllerSet.spec.dbConnectionURL=jdbc:postgresql://postgres/postgresdb?user=postgresadmin&password=admin123"
 ```
 
+### Configuring storage pool creation
+The piraeus operator installed by helm can be used to create storage pools. Creation is under control of the
```
JoelColledge, line 106: I would prefer it we kept the style consistent - that is, with empty lines after the headings and around the code blocks. I know this is nitpicking and I will merge it anyway, but it would be nice.

README.md
```diff
@@ -108,6 +102,64 @@ etcd cluster by adding the following to the Helm install command:
...
+```
+helm install -f <file> charts/piraeus-op
+```
+
+#### After install
+On a cluster with the operator already configured (after helm install)
+
+* Get the current configuration:
+```
+$ kubectl get LinstorNodeSet.piraeus.linbit.com <nodesetname> -o yaml > nodeset.yaml
```
JoelColledge, line 152: Does `kubectl edit` not work here? It makes the instructions shorter and anyone with kubernetes experience will know that they can do it this way if they prefer.

WanzenBug:
Implemented the suggested changes

PR #8, opened 2020-05-07
simplify image selection by using plain strings

WanzenBug:
remove the separation between image repo, name, and tag.
A specific image is identified by a single string.

* Update helm charts
* Update CRDs
* Update go models
* Update go operator


JoelColledge:
I love changes that simplify things and remove code :-)

PR #10, opened 2020-05-07
remove redundant constants

WanzenBug:
All removed constants are required according to the schema,
so they were never used.

From my understanding, these code paths were effectively unreachable, so no user facing changes here.


JoelColledge:
I think it might still be worth leaving a validation check in so that we fail reconciliation if any of the required values are empty. The problem is that we have been "breaking the rules" by changing the CRD schema without introducing a new schema version. When a user upgrades and replaces the CRD, existing CRs may not have the required fields, but the API server won't throw these CRs away, it will probably just leave the fields empty. I'm not totally sure whether this is correct, but I think it would be pragmatic to leave a check until we change the schema version.

Issue #5, opened 2020-04-10
Feature request: Encrypted communication


tbykowsk:
Please consider providing an option to enable encrypted communication (_https_) between Linstor components.
Ideally, this could also include the communication within and to an _etcd_ database cluster, which is used by a Linstor Controller.

JoelColledge:
So there are many parts to this, as you have already indicated:
* Linstor Client - Controller: For this, the user will need access to a certificate to be able to connect. Do you know how this is typically solved? I'm assume the user would be expected to supply a certificate in a secret or something of that nature. The Operator and CSI driver would also need to be supplied with the necessary certificate to be able to connect to the Controller.
* Linstor Controller - Satellite: The operator could configure this without any interaction from the user.
* Linstor Controller - etcd cluster: This probably also requires certificates from the user, because they need to be able to access the etcd cluster directly as well.
* Within etcd cluster: This is outside the scope of the operator itself. The Helm chart should allow it already just using the `auth.peer` options [documented here](https://github.com/bitnami/charts/tree/master/bitnami/etcd), but with the `etcd.` prefix for the keys.

tbykowsk:
In cases when a key is required from the user, a common practice is to create a Certificate Signing Request for this key, using `certificates.k8s.io` API - which provides you with a TLS certificate signed by a Certificate Authority. Then you usually supply the key/certificate to pods using secrets.

Thank you for pointing out that the etcd cluster is an external dependency. Indeed, I see that communication within the cluster might be set separately via `etcd-env-vars` ConfigMap (from the examples). If a Linstor Controller could be configured by the operator to communicate with the etcd using a certificate provided by a secret, then it would be great.

JoelColledge:
@WanzenBug Has implemented a solution for the "Linstor Controller - etcd cluster" part so it should now be possible to encrypt all communication with and within the etcd cluster.

JoelColledge:
#13 implements the "Linstor Controller - Satellite" part.

WanzenBug:
I have a question in regards to this request. @tbykowsk, I hope you can help us answer it!

When you say:
>  enable encrypted communication
does that mean:
1. Components talk via encrypted channels (HTTPS, TLS, etc)
2. Components talk via encrypted channels (HTTPS, TLS, etc), which is **_also_** used for authentication

[...]

PR #11, opened 2020-05-11
enable secure database connection

WanzenBug:
access to the linstor database should be securable.
LINSTOR already supports this. This commit allows to configure LINSTOR to use https when talking to an `etcd` instance.

Partly implements #5


JoelColledge:
Looks great.

PR #13, opened 2020-05-12
secure comms between linstor components

WanzenBug:
Add configuration switches to enable TLS secured comms between
linstor components. This implementations expects that all keys
and certs are already fully functional and converted into the
java keystore format.

Because of a current limitation of LINSTOR, it is not possible to change the keystore password on the controller (it IS possible on nodes). This limitation is currently mentioned in the README and also checked in the operator itself.


README.md
```diff
@@ -56,6 +56,9 @@ The operator can be deployed with Helm v3 chart in /charts as follows:
   * Disable persistence for basic testing. This can be done by adding `--set
     etcd.persistence.enabled=false` to the `helm install` command below.
 
+- If needed, configure secure communication between LINSTOR components.
+  [Read more below](#configuring-secure-communication-between-LINSTOR-components).
```
JoelColledge, line 60: This README is getting too long. Could you create `doc/security.md` or similar and move the encryption docs there?

charts/piraeus/values.yaml
```diff
@@ -18,8 +18,16 @@ operator:
   controllerSet:
     controllerImage: "quay.io/piraeusdatastore/piraeus-server:v1.7.0"
     dbCertSecret: ""
+    sslSecret: ""
+    sslKeystorePassword: "linstor"
+    sslKeyPassword: "linstor"
+    sslTruststorePassword: "linstor"
```
JoelColledge, line 24: You put the SSL keys in an "ssl" subsection in the CRD, which feels neater to me. Why not also do that here?

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -178,6 +178,13 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, fmt.Errorf("CS Reconcile: missing required parameter dbConnectionURL: outdated schema")
 	}
 
+	if pcs.Spec.Ssl != nil {
+		// 2020-05-12 FIXME: Teach LINSTOR how to read these parameters from controller config. Currently these values are added directly to the database on initial migration.
```
JoelColledge, line 182: I wouldn't write FIXME here because this is nothing that can be fixed in this project. Just write "LINSTOR cannot currently ...".

pkg/utils/utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
```
JoelColledge, line 1: Could we use a more descriptive package name than "utils"?

charts/piraeus/values.yaml
WanzenBug: It's recommended by [helm](https://helm.sh/docs/chart_best_practices/values/) because it makes it easier to override "flat" values. However Now thinking about it, in 99% (100%?) of all cases the other `ssl` parameters won't be touched, so it might as well be its own subsection

JoelColledge: If we decide not to make the passwords configurable at all, then this will become irrelevant anyway.

WanzenBug:
> It doesn't really increase the security since we're just writing them into some CRDs which are presumably less well protected than the secrets themselves

This is actually a good point. It still feels strange to "force" users to use a specific password. (actually the java keystore has a similar issue where you can specify a key password separate from the file password, with the restriction that it can't be different from the file password)

For now I would just remove all mentions of a selectable password and force the user to create the keystores with `linstor` as password

README.md
WanzenBug: Replaced this part with a general step to check the security guide

charts/piraeus/values.yaml
WanzenBug: removed, password is always `linstor` now

pkg/utils/utils.go
WanzenBug: removed with some more refactoring

WanzenBug:
@JoelColledge please re-review

JoelColledge:
This feels right to me. The security features are important for some users, but we want to keep them out of the way for people who are just trying it out.

PR #15, opened 2020-05-13
handle ControllerSets that have not been deployed when finalizing

JoelColledge:
If the ControllerSet was never deployed, for instance because there were
no appropriate nodes to deploy it on, then finalization should be a
no-op.

Fixes an issue where finalization failed because the controller could
not be contacted to check for registered nodes.

@WanzenBug - could you check this?


WanzenBug:
Looks good!

Do you know if we currently have any kind of timeout when checking the controller API? For me it looks like in theory we could be waiting indefinitely on a response from the controller....

JoelColledge:
> Do you know if we currently have any kind of timeout when checking the controller API?

Looks like we are using `context.TODO()` all over the place. We should fix this. But this commit doesn't make the situation any worse, so I'll merge it.

PR #17, opened 2020-05-14
crds: change status fields to use lists instead of maps

WanzenBug:
Maps are not recommended in k8s apis [SRC#1].
operator-sdk refuses to automatically generate crds.

this commit changes the maps in status fields to lists, regenrates generated code.

[SRC#1]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps


pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -191,10 +191,6 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
 		"spec":      fmt.Sprintf("%+v", pcs.Spec),
 	}).Debug("found LinstorControllerSet")
 
-	if pcs.Status.SatelliteStatuses == nil {
-		pcs.Status.SatelliteStatuses = make(map[string]*piraeusv1alpha1.SatelliteStatus)
-	}
```
JoelColledge, line 196: I think we should initialize all lists in the CR structs when we create/get them. At the moment, if there is no controller pod (e.g. because the priority class doesn't exist), then you end up with
```
status:
  ControllerStatus:
    nodeName: ""
    registeredOnController: false
  SatelliteStatuses: null
  errors:
  - 'CS Reconcile: waiting until there is exactly one controller pod, found 0 pods
    instead'
```

We want to avoid `null` for lists in the serialized representation, so we should use empty slices instead of `nil`. This agrees with https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -410,11 +407,6 @@ func (r *ReconcileLinstorNodeSet) reconcileSatNodes(pns *piraeusv1alpha1.Linstor
 		if out.err != nil {
 			compoundError = append(compoundError, out.err)
 		}
-
-		// Guard against empty statuses.
-		if out.sat == nil || out.sat.NodeName != "" {
-			pns.Status.SatelliteStatuses[out.sat.NodeName] = out.sat
```
JoelColledge, line 416: Looks like `out.sat` is no longer necessary, so `satelliteStatusOut` only needs to be a channel of errors.

I have the feeling that the whole `satelliteStatusIn`, `satelliteStatusOut` logic is unnecessarily complicated. Doing this concurrently with a concurrent operator limit is reasonable, but I don't see why it needs to be split into 2 stages. But that's a different issue, apart from the potential to remove `type satStat` now.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -489,9 +482,7 @@ func (r *ReconcileLinstorNodeSet) reconcileStoragePoolsOnNode(sat *piraeusv1alph
 		}).Debug("NS reconcileStoragePoolsOnNode: found storage pool")
 
 		// Guard against empty statuses.
-		if status == nil || status.Name != "" {
```
JoelColledge, line 492: Have you checked whether/how this might occur? One assumes that there was some reason for it in the first place.

WanzenBug: Its a bit strange, because in case the `status == nil` condition holds, this will lead to a nil dereference.
Also status is created in line 485 above, which is ensures that status is never nil.

PR #18, opened 2020-05-15
csi deploy in go: translate helm yamls to go operator

WanzenBug:
CSI containers are now also deployed via go operator. This has some
implications:

* There is a CRD `LinstorCsiDriver`which controls the creation of CSI
  infrastructure
* Its possible to deploy linstor without CSI drivers
* In theory, its possible to deploy CSI without linstor (or external
  linstor controller)
* The operator might need to create service accounts, cluster roles,
  etc. For it to be able to do that, it needs to have these permissions
  itself.


WanzenBug:
This MR is nearly done:
* I added very basic unit tests that just ensure that the expected resources are created (actually, only that the resources with a given name exist, further checks are not implemented)
* I checked that the correct watches for the operator are set and a basic status field exists for the CRD. Because of the watches its not so easy to implement a "update" logic for existing resources (every update triggers a `.Reconcile()` which leads to an endless loop of reconciliation). For now, its probably fine to just use a "create if not exist" logic.

TODO:
* I want to look through the required roles of the operator and simplify the helm templates required for operator roles

WanzenBug:
I moved all the different roles etc into a single helm template. That way I find it a bit easier to navigate between `serviceaccount`-> `binding` ->`role`

@JoelColledge could you please take a look

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -0,0 +1,878 @@
...
+									Value: "/csi/csi.sock",
+								},
+								{
+									Name:  "DRIVER_REG_SOCK_PATH",
+									Value: "/var/lib/kubelet/plugins/linstor.csi.linbit.com/csi.sock",
+								},
+								{
+									Name: "KUBE_NODE_NAME",
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
```
JoelColledge, line 522: Too much nesting. Maybe move the `Container` construction into its own function?

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -0,0 +1,878 @@
...
+							Args: []string{
+								"--provisioner=linstor.csi.linbit.com",
+								"--csi-address=$(ADDRESS)",
+								"--v=5",
+								"--feature-gates=Topology=false",
+								"--connection-timeout=4m",
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name:  "ADDRESS",
```
JoelColledge, line 665: Too much nesting, as above.

pkg/controller/linstorcsidriver/linstorcsidriver_controller_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	DefaultPriorityClass = schedv1.PriorityClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo-csi-priority-class",
+			Namespace: "bar",
+		},
+	}
+)
+
+func TestReconcileLinstorCSIDriver_Reconcile(t *testing.T) {
```
JoelColledge, line 54: This is really cool. We should add tests for the other controllers and whenever we fix bugs.

charts/piraeus/values.yaml
```diff
@@ -11,7 +11,12 @@ etcd:
     replicaCount: 3
   envVarsConfigMap: "etcd-env-vars"
 csi:
-  image: "quay.io/piraeusdatastore/piraeus-csi:v0.7.4"
+  enabled: true
+  pluginImage: "quay.io/piraeusdatastore/piraeus-csi:v0.7.4"
+  imagePullSecret: ""
+  nodeServiceAccountName: ""
+  controllerServiceAccountName: ""
+  priorityClass: ""
```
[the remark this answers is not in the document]
WanzenBug: On one hand, I was a bit hesitant to add so many new permissions to the operator without an "escape hatch". On the other hand, I have not really invested in documenting this. In practice this probably means the options won't be used anyways.

I think I will remove them

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -0,0 +1,878 @@
...
+// and what is in the LinstorCSIDriver.Spec
+func (r *ReconcileLinstorCSIDriver) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := logrus.WithFields(logrus.Fields{
+		"requestName":      request.Name,
+		"requestNamespace": request.Namespace,
+	})
+	reqLogger.Info("Reconciling LinstorCSIDriver")
+
+	// Fetch the LinstorCSIDriver instance
+	cr := &piraeusv1alpha1.LinstorCSIDriver{}
```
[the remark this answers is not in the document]
WanzenBug: updated

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -0,0 +1,878 @@
...
+	if specError != nil {
+		cr.Status.Errors = []string{specError.Error()}
+	} else {
+		cr.Status.Errors = []string{}
+	}
+
+	cr.Status.NodeReady = nodeReady
+	cr.Status.ControllerReady = controllerReady
+
+	err = r.client.Status().Update(ctx, cr)
```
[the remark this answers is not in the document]
WanzenBug: added the additional logging

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
WanzenBug: Split up the literal into smaller chunks with (hopefully) descriptive names. I kept them all in the same function as they are closely related anyways

WanzenBug:
@JoelColledge all feedback is now integrated and commits are squashed

PR #20, opened 2020-05-19
etcd-auth: add option to authenticate on etcd with certificates

WanzenBug:
etcd supports client authentication when HTTPS is enabled. The client is
expected to present a valid certificate, where CN is set to the user name.

This commit enables authenticating LINSTOR against a etcd instance using
these client certificates.

Addresses part of #5

### DIFF
--- a/charts/piraeus/crds/operator-controllerset-crd.yaml
+++ b/charts/piraeus/crds/operator-controllerset-crd.yaml
@@ -44,6 +44,11 @@ spec:
                 PEM format
               type: string
               nullable: true
+            dbUseClientCert:
+              description: Use a TLS client certificate for authentication with the database (etcd). If set to true,
+                `dbCertSecret` must be set and contain two additional entries "client.cert" (PEM encoded)
+                and "client.key" (PKCS8 encoded, without passphrase).
+              type: boolean
             luksSecret:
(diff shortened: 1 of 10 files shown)


charts/piraeus/crds/operator-controllerset-crd.yaml
```diff
@@ -44,6 +44,11 @@ spec:
                 PEM format
               type: string
               nullable: true
+            dbPresentClientCert:
+              description: Present a TLS client certificate for authentication with the database (etcd). If set to true,
+                `dbCertSecret` must be set and contain two additional entries "client.cert" and "client.key" in PEM
+                format.
```
JoelColledge, line 50: Your docs say that one of them should be in PKCS8 format.

charts/piraeus/crds/operator-controllerset-crd.yaml
```diff
@@ -44,6 +44,11 @@ spec:
                 PEM format
               type: string
               nullable: true
+            dbPresentClientCert:
```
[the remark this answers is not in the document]
WanzenBug: `s/PresentClientCert/UseClientCert/g` done :-)

charts/piraeus/crds/operator-controllerset-crd.yaml
WanzenBug: Updated

Issue #3, opened 2020-04-10
Feature request: NVMe layer support


tbykowsk:
Please consider providing an option to support the NVMe layer for Linstor Satellites.
For example, a Satellite's pod might access modules required for the NVMe layer from a host, or maybe modules could be provided as a sidecar container.

JoelColledge:
Thanks for the request, it seems reasonable to me. For initial NVMe layer support I would assume that any necessary setup that can be done on the host, has already been done so. For instance, the installation of any necessary kernel modules. Then the satellite will just require some additional userspace tools and maybe some more mounts.

tbykowsk:
Yes, for the initial support using modules from the host would be perfect. Thank you for a quick response!

JoelColledge:
The NVMe tools are being added to the satellite images and that should be sufficient for this to work.

* The Piraeus Dockerfile has been updated (https://github.com/piraeusdatastore/piraeus/commit/e6c940ac8e4dc8cf388893476ffc713f30486112) but the image on quay.io has not yet been updated
* The drbd.io image has been updated. `drbd.io/linstor-satellite:v1.6.2.1` contains the tools

Please close the issue once it is working for your requirements.

tbykowsk:
Thanks, I've checked the new image, but I got an error.
It seems that a pod with the satellite does not have access to `/sys` directory on a host:
```
sh-4.2$ linstor resource create --layer-list nvme,storage --storage-pool testpool storage-1-1 res3
SUCCESS:
    Successfully set property key(s): StorPoolName
SUCCESS:
Description:
    New resource 'res3' on node 'storage-1-1' registered.
Details:
    Resource 'res3' on node 'storage-1-1' UUID is: dffab0cd-ccea-415a-8cbc-3c091056dcdc
SUCCESS:
Description:
    Volume with number '0' on resource 'res3' on node 'storage-1-1' successfully registered
Details:
    Volume UUID is: fe19865f-fc2c-4214-8bd2-83ffdebcb79c
ERROR:
    (Node: 'storage-1-1') Failed to configure NVMe target!
Show reports:
    linstor error-reports show 5EB92A2E-9C9F2-000007
sh-4.2$ linstor error-reports show 5EB92A2E-9C9F2-000007
ERROR REPORT 5EB92A2E-9C9F2-000007

============================================================

Application:                        LINBIT? LINSTOR
Module:                             Satellite
Version:                            1.7.0
[...]

PR #21, opened 2020-05-19
linstor: bind-mount /sys/ in satellites for NVMe support

WanzenBug:
linstor satellites supporting NVMe need access to:
* `/sys/kernel/config/nvmet/`
* `/sys/devices/virtual/nvme-fabrics/ctl/nvme`
While the second entry is mounted from `sysfs`, which is already
present in the container, the first entry is part of the `configfs`
mount `/sys/kernel/config`, which is not mounted by default.

With this change, linstor-satellite will bind-mount `/sys` from the host.
As long as the host has the required kernel modules loaded for NVMe,
linstor will be able to control the NVMe layer.

This is needed for #3

### DIFF
--- a/pkg/controller/linstornodeset/linstornodeset_controller.go
+++ b/pkg/controller/linstornodeset/linstornodeset_controller.go
@@ -551,6 +551,10 @@ func newDaemonSetforPNS(pns *piraeusv1alpha1.LinstorNodeSet, config *corev1.Conf
 									MountPath: kubeSpec.DevDir,
 								},
 								{
+									Name: kubeSpec.SysDirName,
+									MountPath: kubeSpec.SysDir,
+								},
+								{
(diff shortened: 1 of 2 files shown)


WanzenBug:
For completeness, while I don't have a NVMe setup, I can still load the nvmet module on a host and check:
```
$ docker inspect 1fbf6a9edd07
...
        "HostConfig": {                                     
            "Binds": [                                      
                "/sys/:/sys/",                              
                "/lib/modules/:/lib/modules/:rshared",      
...

$ kubectl exec -it piraeus-op-ns-node-8zz9d find /sys/kernel/config/nvmet/
/sys/kernel/config/nvmet/
/sys/kernel/config/nvmet/hosts
/sys/kernel/config/nvmet/ports
/sys/kernel/config/nvmet/subsystems
```

PR #23, opened 2020-05-19
Make CSI images configurable

WanzenBug:
The various sidecar images are now configurable via the LinstorCSIDriver resource.
To stay backwards compatible, the values are not required and the resource will automatically
be updated with the defaults

### DIFF
--- a/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
+++ b/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
@@ -133,6 +133,11 @@ func (r *ReconcileLinstorCSIDriver) Reconcile(request reconcile.Request) (reconc
 
 	ctx, _ := context.WithTimeout(context.Background(), 30*time.Second)
 
+	resourceErr := r.reconcileResource(ctx, csiResource)
+	if resourceErr != nil {
+		return reconcile.Result{}, resourceErr
+	}
+
 	specErr := r.reconcileSpec(ctx, csiResource)
 
 	statusErr := r.reconcileStatus(ctx, csiResource, specErr)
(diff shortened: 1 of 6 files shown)


pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -143,6 +148,37 @@ func (r *ReconcileLinstorCSIDriver) Reconcile(request reconcile.Request) (reconc
 	return reconcile.Result{}, statusErr
 }
 
+func (r *ReconcileLinstorCSIDriver) reconcileOldSpec(ctx context.Context, csiResource *piraeusv1alpha1.LinstorCSIDriver) error {
+	changed := false
+
+	// ac78c93 2020-05-19 Add options for CSI sidecar container images
```
JoelColledge, line 154: What's the purpose of this comment? We can assume that the reader knows how to use git blame if they are interested in the history. Also, the commit hash will be incorrect after a rebase, which will happen because we use the "rebase and merge" strategy.

WanzenBug: The main purpose of this comment was to signal _why_ these changes/updates are needed. I envisioned this function to be kind of "append only", so that we have a clean upgrade path for older resources.

You are right that the commit hash would change after rebasing this branch, however as long as it is up-to-date on merge into master, it should never ever change. In my experience `git blame` often needs some manual review because someone refactors whole parts of files and so changes need to be excluded, traced to older files, etc...

However, you are probably right that it is a bit of a brittle setup, I will rework it

WanzenBug: Replaced with some verbose logging code, which I think can serve the same purpose

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -143,6 +148,47 @@ func (r *ReconcileLinstorCSIDriver) Reconcile(request reconcile.Request) (reconc
 	return reconcile.Result{}, statusErr
 }
 
+func (r *ReconcileLinstorCSIDriver) reconcileOldSpec(ctx context.Context, csiResource *piraeusv1alpha1.LinstorCSIDriver) error {
+	logger := logrus.WithFields(logrus.Fields{
+		"Name":      csiResource.Name,
+		"Namespace": csiResource.Namespace,
+		"Op":        "reconcileOldSpec",
+	})
+	logger.Info("performing potential upgrades on old resources")
```
JoelColledge, line 157: Doesn't this run every time you install CSI with Helm and our default Helm values? The default image values are empty and this is the only place which fills them with defaults. So it's not specific to upgrades.

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
JoelColledge: Just a technical detail:

> as long as it is up-to-date on merge into master, it should never ever change

That holds for how GitLab does fast-forward merges, but GitHub rewrites the committer, so the hashes change.

WanzenBug: Huh. Never noticed that. But it does make sense

PR #24, opened 2020-05-22
repo: add github actions CI

WanzenBug:
Create infrastructure for https://pre-commit.com/ and checking via
github actions.


JoelColledge:
Interesting approach. Do you have experience with `pre-commit`?

WanzenBug:
I've used it previously for some Python projects. It makes it quite easy to integraten"standard" steps like running `go test` with steps like recreating crds from models, fixing whitespace etc.

WanzenBug:
The infrastructure is almost there now. However, since we haven't been running golangci-lint for some time, there are a lot of changes to make. I would wait with fixing them until after the linstor-client HTTPS feature is implemented. 

I used my [personal fork to test the CI setup](https://github.com/WanzenBug/piraeus-operator/pull/2) (which reports a lot more files as changed, so I get a lot of linter warnings) 
image

WanzenBug:
This is ready for review. I did fix all the issues reported by golangci-lint and gofumpt, they can be fixed as they come up when changing files.

Checks are all green now: https://github.com/WanzenBug/piraeus-operator/pull/2/checks

WanzenBug:
@JoelColledge could you please take a look?

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+          key: ${{ runner.os }}-build-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-build-
+      - name: Build with operator-sdk
+        run: operator-sdk build quay.io/piraeusdatastore/piraeus-operator:latest
+      - name: Push docker image
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
JoelColledge, line 40: Do these secrets already exist?

WanzenBug: No :sweat_smile:
Automatic push is the only thing I could not really test in my fork (docker build works fine)

JoelColledge: @rck Could you configure secrets for pushing to `quay.io`? I don't have the permissions.

.github/workflows/golangci.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v1
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
```
JoelColledge, line 18: No need to copy the documentation of the action in here

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2.0.3
+      - uses: actions/cache@v1
+        with:
+          path: ~/operator-sdk/bin/
+          key: operator-sdk-${{ env.OPERATOR_SDK_VERSION }}
```
JoelColledge, line 19: Where does `OPERATOR_SDK_VERSION` actually come from? I see that it is set explicitly in `pre-commit.yml`, but what about here?

.github/workflows/golangci.yml
```diff
@@ -0,0 +1,28 @@
+name: golangci-lint
```
JoelColledge, line 1: Can we not run this as part of precommit?

WanzenBug: We can, the advantage of this approach is that the github action will automatically create notes on the relevant diffs

.github/workflows/build.yml
WanzenBug: copy&paste error

.github/workflows/golangci.yml
JoelColledge: OK, that makes it worthwhile

.github/workflows/golangci.yml
WanzenBug: Removed

.github/workflows/build.yml
WanzenBug: fixed

JoelColledge:
Any idea why the checks aren't running for this PR?

WanzenBug:
I think because I'm "external" for the purpose of this repo, so I can't create any workflows.

WanzenBug:
In my fork they are running as expected: https://github.com/WanzenBug/piraeus-operator/pull/2/checks

charts/piraeus/crds/piraeus.linbit.com_linstorcontrollersets_crd.yaml
```diff
@@ -32,47 +32,33 @@ spec:
...
-                controller/server container
-              type: string
-            linstorHttpsControllerSecret:
-              description: Name of the secret containing the java keystore (`keystore.jks`)
-                used to enable HTTPS on the controller. The controller will create
-                a secured https endpoint on port 3371 with the key stored in `keystore.jks`.
-                The keystore must be secured using the passphrase "linstor". Also needs to
-                contain a truststore `truststore.jks`, which will be used to authenticate
-                clients.
+              description: DBCertSecret is the name of the kubernetes secret that
```
JoelColledge, line 57: This description doesn't belong with the key `drbdRepoCred`.

I will stop reviewing here and let you reconcile the differences between the CRDs are they were and these newly generated ones. This yaml differ https://eloquent-hodgkin-f52b42.netlify.app/ seems to work pretty well. The result doesn't have to be identical to the original, but we should make sure we're not making anything worse.

.github/workflows/build.yml
rck: I will take care of that.

charts/piraeus/crds/piraeus.linbit.com_linstorcontrollersets_crd.yaml
WanzenBug: fixed, result is identical for validation purpose. Updated the descriptions with whatever seemed more appropriate from the old CRDs or go comments

.github/workflows/build.yml
JoelColledge: The secrets now exist

PR #25, opened 2020-05-25
linstor: enable https for rest interface

WanzenBug:
* Enable HTTPS if secret with server key is given controllerset.
* Enable HTTPS client auth via certificates if option
  `linstorHttpsClientUseClientCert` enabled
* Automatically switch between http and https communication for internal
  API calls

TODO/Known Issues:
* [x] Duplicated code between nodeSet/controllerSet (should be merged in any case?)
* [x] Duplication in helm setup (can be removed without huge refactoring)
* [x] Docs on how to create certificates / expected keys in secrets
* [x] Duplicate services for controller? why do we have piraeus-op-ns and piraeus-op-cs both pointing at the controller pod?


JoelColledge:
> Duplicate services for controller? why do we have piraeus-op-ns and piraeus-op-cs both pointing at the controller pod?

I think some people have the expectation that it is a cluster system so I should be able to send my commands to any node. This is roughly how Pacemaker works. The old DRBD Manage was similar too, I believe. We should just remove the piraeus-op-ns service. I think the fact that you send your commands to the controller fits in with the Kubernetes way of thinking fairly well.

charts/piraeus/values.yaml
```diff
@@ -26,9 +26,14 @@ operator:
...
+    linstorHttpsClientSecret: ""
+    linstorHttpsUseClientCert: false
   nodeSet:
     drbdKernelModuleInjectionMode: Compile
     satelliteImage: "quay.io/piraeusdatastore/piraeus-server:v1.7.0"
     kernelModImage: "quay.io/piraeusdatastore/drbd9-bionic:v9.0.22"
     storagePools: null
     sslSecret: ""
+    linstorHttpsUseClientCert: false
+    linstorHttpsClientSecret: ""
```
JoelColledge, line 39: I don't think it makes much sense to duplicate these keys in the Helm values. Either we just use the ones specified in `controllerSet` everywhere, or we put them in their own section.

charts/piraeus/crds/piraeus.linbit.com_linstorcsidrivers_crd.yaml
```diff
@@ -57,6 +57,17 @@ spec:
             linstorControllerAddress:
               description: Cluster address of the linstor controller
               type: string
+            linstorHttpsClientSecret:
```
JoelColledge, line 60: I don't see this being set anywhere in the Helm chart yet.

WanzenBug:
There are some unexpected problems with the current draft regarding the CSI infrastructure. The current version of the CSI plugin only allows enabling HTTPS if [TLS user authentication is also used](https://github.com/piraeusdatastore/linstor-csi/blob/v0.8.1/cmd/linstor-csi/linstor-csi.go#L73)

This code is actually also duplicated in [the golinstor client](https://github.com/LINBIT/golinstor/blob/v0.26.0/client/client.go#L146)

For now, I will make using client certificates mandatory if HTTPS is enabled

charts/piraeus/values.yaml
WanzenBug: Moved to the top level

charts/piraeus/crds/piraeus.linbit.com_linstorcsidrivers_crd.yaml
WanzenBug: Fixed

WanzenBug:
@JoelColledge ready for review

charts/piraeus/crds/operator-controllerset-crd.yaml
```diff
@@ -66,6 +66,21 @@ spec:
...
+            linstorHttpsControllerSecret:
+              description: Name of the secret containing the java keystore (`keystore.jks`)
+                used to enable HTTPS on the controller. The controller will create
+                a secured https endpoint on port 3371 with the key stored in `keystore.jks`.
+                The keystore must be secured using the passphrase "linstor". Also needs to
+                contain a truststore `truststore.jks`, which will be used to authenticate
+                clients.
+              type: string
+            linstorHttpsClientSecret:
+              description: Name of the secret containing * the root certificate (`ca.pem`)
```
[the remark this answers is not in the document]
WanzenBug: Seems to caused by `operator-sdk generate crds`. I will try to get it to preserve formatting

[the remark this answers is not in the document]
JoelColledge: If that doesn't work, we could just use a style which is readable without line breaks. E.g. `(a) ... (b) ...`.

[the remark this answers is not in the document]
WanzenBug: Updated with your suggestion

WanzenBug:
For completeness: This should be the last step for #5

PR #31, opened 2020-06-04
allow customizing linstor controller endpoint

WanzenBug:
Adds an optional (for now) "controllerEndpoint" field to all resources. If
not given, the default value will be filled and the resource updated.

This allows pointing nodesets and CSI components at different controllers if
needed. Also determening the controller URL is made much simpler.

### DIFF
--- a/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
+++ b/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
@@ -199,13 +199,8 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
 		return secret.Data, nil
 	}
 
-	tlsConfig, err := lc.ApiResourceAsTlsConfig(&pcs.Spec.LinstorClientConfig, getSecret)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	serviceName := types.NamespacedName{Name: pcs.Name, Namespace: pcs.Namespace}
-	r.linstorClient, err = lc.NewHighLevelLinstorClientForServiceName(serviceName, tlsConfig)
+	endpoint := expectedEndpoint(pcs)
+	r.linstorClient, err = lc.NewHighLevelLinstorClientFromConfig(endpoint, &pcs.Spec.LinstorClientConfig, getSecret)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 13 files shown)


charts/piraeus/crds/piraeus.linbit.com_linstorcontrollersets_crd.yaml
```diff
@@ -32,6 +32,10 @@ spec:
         spec:
           description: LinstorControllerSetSpec defines the desired state of LinstorControllerSet
           properties:
+            controllerEndpoint:
+              description: Cluster URL of the linstor controller. If not set, will
+                be determined from the current resource name.
+              type: string
```
WanzenBug, line 38: It's pretty redundant to have this in the controller set. However this way it is consistent across all 3 controllers. Users probably never need to set this value in any case.

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -338,6 +337,37 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
 	return reconcile.Result{RequeueAfter: time.Minute * 1}, compoundError
 }
 
+func (r *ReconcileLinstorControllerSet) reconcileResource(ctx context.Context, pcs *piraeusv1alpha1.LinstorControllerSet) error {
+	logger := logrus.WithFields(logrus.Fields{
+		"Name":      pcs.Name,
+		"Namespace": pcs.Namespace,
+		"Op":        "reconcileResource",
+	})
+	logger.Info("performing upgrades and fill defaults in resource")
```
JoelColledge, line 346: Maybe we should put these logs at the debug level, and just have info logging when values are actually set (inside the if) because that is the only case that is normally interesting.

WanzenBug:
I agree that its not ideal, however I still went with it because:
* its part of the same shared code for configuring 3 different clients
  - linstor-cli (controller/nodes)
  - go-client (csi/node/controller operator)
  - csi client (csi)
* minor: if HTTPS is enabled, its a good way to ensure that the correct server name is used

I don't think many users will get confused by a parameter they probably never have to set.
Instead we should just make a simple sanity check, to make sure the URL matches the generated one

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
WanzenBug: implemented on all resource upgrade/fill paths

charts/piraeus/crds/piraeus.linbit.com_linstorcontrollersets_crd.yaml
JoelColledge: It is not great making our API messy for the sake of internal implementation details. Also, I don't think it makes that much difference to the code. We should just be able to take `ControllerEndpoint` out of `LinstorClientConfig` and put it directly in the CSI and node structs. Then we just pass it as an additional argument to `NewHighLevelLinstorClientFromConfig` and `NewClientConfigForAPIResource`.

It would trouble me less if it were documented as "internal, do not set" in the ControllerSet CRD, but I can't see how we can manage that without breaking the generation of CRDs from code.

WanzenBug: Done

PR #32, opened 2020-06-04
create rbac rules outside of operator

WanzenBug:
RBAC can be handled by all recommended deployment tools, namely:
* helm
* OLM
This commit removes all RBAC related rules from the CSI operator and
moves creation to Helm.

### DIFF
--- a/charts/piraeus/templates/csi-controller-rbac.yml
+++ b/charts/piraeus/templates/csi-controller-rbac.yml
@@ -1,0 +1,121 @@
+# This YAML file contains all RBAC objects that are necessary to run a
+# CSI controller pod
+#
+# This file was manually merged from the following sources:
+# external-attacher v2.2.0: https://raw.githubusercontent.com/kubernetes-csi/external-attacher/v2.2.0/deploy/kubernetes/rbac.yaml
+# external-provisioner v1.6.0: https://raw.githubusercontent.com/kubernetes-csi/external-provisioner/v1.6.0/deploy/kubernetes/rbac.yaml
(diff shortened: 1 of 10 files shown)


charts/piraeus/templates/csi-controller-rbac.yml
```diff
@@ -0,0 +1,121 @@
...
+roleRef:
+  kind: ClusterRole
+  name: external-provisioner-runner
+  apiGroup: rbac.authorization.k8s.io
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-snapshotter-runner
```
JoelColledge, line 94: What actually happens when someone want to install 2 CSI drivers in their cluster? Can one deploy the generic CSI components twice?

Here we are creating a non-namespaced resource with a generic name. This could clash with the deployment of a different driver, if it works similarly. Of course, it the resource may well be identical, but it could be different if the other driver deploys a different CSI version, for instance. That made me wonder whether the CSI support in the operator is compatible with this situation at all.

Probably not relevant to this PR, unless we want to change the names to avoid such clashes.

WanzenBug: > Can one deploy the generic CSI components twice
Yes, as the components themselves are namespaced.

I think its fine to have this one non-namespaced resource with generic name. This role can be shared between all CSI deployments, as its just the requirement for the generic CSI snapshotter. The serviceaccounts are namespaced, and the role bindings are able to list multiple service accounts

PR #34, opened 2020-06-08
automatically create storage from available devices

WanzenBug:
Add a new option to automatically set up storage pools. Using
LINSTOR's create-physical-storage command, a storage pool of the
given type is automatically created on all devices.


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+* Automatic storage pool creation via `automaticStorageSetupType` on `LinstorNodeSet`. If this option is set, LINSTOR
+  will create a storage pool based on all available devices on a node. This option should only ever be used in a test
+  environment.
```
JoelColledge, line 13: I'm not sure we need this warning. The process should be fairly safe.

charts/piraeus/crds/piraeus.linbit.com_linstornodesets_crd.yaml
```diff
@@ -31,6 +31,20 @@ spec:
         spec:
           description: LinstorNodeSetSpec defines the desired state of LinstorNodeSet
           properties:
+            automaticStorageSetupType:
```
JoelColledge, line 34: A bit shorter: `automaticStorageType`?

charts/piraeus/values.yaml
```diff
@@ -35,3 +35,4 @@ operator:
     kernelModImage: "quay.io/piraeusdatastore/drbd9-bionic:v9.0.22"
     storagePools: null
     sslSecret: ""
+    automaticStorageSetupType: None
```
JoelColledge, line 38: I think we should switch this on by default to reduce the number of steps required for someone trying out the operator to get to a working system. Defaut `LVMTHIN`.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -136,9 +138,11 @@ func newCompoundErrorMsg(errs []error) []string {
 // that doesn't make a lot of sense to put elsewhere, so don't lint it for cyclomatic complexity.
 // nolint:gocyclo
 func (r *ReconcileLinstorNodeSet) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx, _ := context.WithTimeout(context.Background(), 1*time.Minute)
+
 	// Fetch the LinstorNodeSet instance
 	pns := &piraeusv1alpha1.LinstorNodeSet{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, pns)
+	err := r.client.Get(ctx, request.NamespacedName, pns)
```
JoelColledge, line 146: Great to remove `context.TODO` :smiley:

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -329,180 +333,260 @@ func (r *ReconcileLinstorNodeSet) reconcileResource(ctx context.Context, pns *pi
...
-		wg.Wait()
-		close(satelliteStatusOut)
-	}()
-
-	var compoundError []error
-
-	for err := range satelliteStatusOut {
+	logger.Debug("start collecting per-node reconciliation results")
+	registerErrors := make([]error, 0)
+	for range pods {
```
JoelColledge, line 397: Maybe worth adding a comment here that we are collecting the number of outputs equal to the number of pods, rather than precisely the outputs for each pod. Otherwise someone will later add logic depending on the pod and they will be mixed up.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -69,6 +68,9 @@ var log = logrus.WithFields(logrus.Fields{
 // to them.
 const linstorNodeFinalizer = "finalizer.linstor-node.linbit.com"
 
+// Default value for AutomaticStorageSetupType. If set, no automatic setup of storage devices happens.
+const defaultAutomaticStorageSetupType = "None"
```
JoelColledge, line 72: I would call this `automaticStorageTypeNone` because isn't just the default; it has a special meaning when this value is used.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -329,180 +334,283 @@ func (r *ReconcileLinstorNodeSet) reconcileResource(ctx context.Context, pns *pi
...
+		"Namespace": nodeSet.Namespace,
+		"Pod":       pod.Name,
+		"Op":        "reconcileAutomaticDeviceSetup",
+	})
+	// No setup required
+	if nodeSet.Spec.AutomaticStorageSetupType == defaultAutomaticStorageSetupType {
+		return nil
+	}
+
+	logger.Warn("Automatic device set up enabled. This is a potentially dangerous operation!")
```
JoelColledge, line 443: Is it really that dangerous? Logging a warning for every node for every reconciliation run is very loud.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -329,180 +334,283 @@ func (r *ReconcileLinstorNodeSet) reconcileResource(ctx context.Context, pns *pi
...
+	// LINSTOR does not like an empty device path list
+	if len(paths) == 0 {
+		logger.Debug("no devices left to set up")
+		return nil
+	}
+
+	logger.WithField("devices", paths).Debug("use devices for storage pool")
+
+	return r.linstorClient.Nodes.CreateDevicePool(ctx, pod.Spec.NodeName, lapi.PhysicalStorageCreate{
+		DevicePaths: paths,
```
JoelColledge, line 474: We may actually want one storage pool per device rather than one for all devices. It is a bit awkward.

One pool per device advantages:
* Better availability - if one device goes down fewer volumes affected
* More flexibility - we don't know if we are combining completely different types of devices here

One for all devices advantages:
* More efficient use of space because we don't have to have space on a single device for a volume

We should get other opinions on this.

WanzenBug:
Made the requested changes:
* Added docs
* Removed excessive warnings
* Changed storage pools to be per-device-name
* Renamed some things
* Extracted parallel registration code into method: this makes it easier to ensure every exit point sends the result via the aggregation channel

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -329,180 +335,301 @@ func (r *ReconcileLinstorNodeSet) reconcileResource(ctx context.Context, pns *pi
...
-		log.Debug("NS reconcileSatNodes: reconciling node")
+	registerErrors := make([]error, 0)
 
-		sat := &piraeusv1alpha1.SatelliteStatus{
-			NodeStatus:          piraeusv1alpha1.NodeStatus{NodeName: pod.Spec.NodeName},
-			StoragePoolStatuses: make([]*piraeusv1alpha1.StoragePoolStatus, 0),
+	// Note: We collect over the _number_ of pods, not the pods it self. There is no guarantee on the order in which
+	// results for each pod are received.
+	for range pods {
+		err := <-output
```
JoelColledge, line 369: We now have stable status ordering, but the errors are still mixed. Could we order them too? Maybe a slice of channels, one for each pod, and then wait for each of them in turn?

doc/storage.md
```diff
@@ -0,0 +1,77 @@
...
+
+The piraeus operator can automate some basic storage set up for LINSTOR. This includes:
+
+* Preparing physical devices. [Read LINSTOR guide](https://www.linbit.com/drbd-user-guide/linstor-guide-1_0-en/#s-a_storage_pool_per_backend_device)
+* Creating storage pools. [Read LINSTOR guide](https://www.linbit.com/drbd-user-guide/linstor-guide-1_0-en/#s-storage_pools)
+
+## Preparing physical devices
+
+The piraeus operator can automatically configure devices on satellites. Depending on the value of
+`operator.nodeSet.automaticStorageType`, all _empty_ devices will be added to a storage pool (one pool per device name,
```
JoelColledge, line 11: Mention that the size must be > 1GiB. It is really `>` and not `>=`, oddly enough.

JoelColledge:
I've left a couple of new comments. Otherwise it's looking good.

PR #37, opened 2020-06-12
add chart to deploy cluster snapshot controller+CRDs

WanzenBug:
Automate the steps described in
> https://github.com/kubernetes-csi/external-snapshotter#usage

to add snapshotting capabilities if k8s vendor did not deploy their own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ The operator can be deployed with Helm v3 chart in /charts as follows:
 
 - Read the [guide on securing the deployment](doc/security.md) and configure as needed.
 
+- Read up on [volume snapshots](#adding-snapshot-support) and configure as needed.
+
 - Finally create a Helm deployment named `piraeus-op` that will set up
   everything.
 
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -110,6 +112,22 @@ etcd cluster by adding the following to the Helm install command:
...
 ```
 
+### Adding snapshot support
+
+LINSTOR supports the volume snapshot feature, which is currently in beta. To use it, you need to install a cluster wide
+snapshot controller. This is done either by the cluster provider, or you can use the piraeus chart.
+
+By default, the piraeus chart will install its own snapshot controller. This can lead to conflict in some cases:
+
+* the cluster already has a snapshot controller
```
JoelColledge, line 122: Do you know how widespread this is? I'm just wondering if our default should be to _not_ install the snapshotter, because that is the safer option and we want to make it as easy as possible to try out the operator.

WanzenBug: Honestly, I have no idea :sweat_smile: 

I guess it depends on how many clusters have multiple CSI drivers configured (which I think is not that common)

PR #38, opened 2020-06-15
requeue reconcile loop at expected wait points

WanzenBug:
At some points we need to wait for specific resource to come online.
* Pods need to be scheduled
* Services need to come online (and responsive)

This change adds a new TemporaryError wrapper around a standard error.
It can be converted to a reconcile result, that automatically requeues
the reconcile loop after a given duration elapsed. We currently use it in:

* the controller, waiting until the pod is running (_NOT_ ready)
* the controller, waiting until the controller responds
* the nodeset, waiting until the controller is reachable
* the nodeset, waiting until the node is online in the controller

A few helper methods are also added, to deal with mixed temporary and
"normal" errors. A temporary error takes precedence over normal errors,
i.e. if in a collection of errors a temporary one exists, the combined
return value with be (`<requeue>`, `<no-error>`)

### DIFF
--- a/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
+++ b/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
@@ -25,9 +25,9 @@ import (
 	"time"
 
 	lapi "github.com/LINBIT/golinstor/client"
-
 	piraeusv1alpha1 "github.com/piraeusdatastore/piraeus-operator/pkg/apis/piraeus/v1alpha1"
 	mdutil "github.com/piraeusdatastore/piraeus-operator/pkg/k8s/metadata/util"
+	"github.com/piraeusdatastore/piraeus-operator/pkg/k8s/reconcileutil"
 	kubeSpec "github.com/piraeusdatastore/piraeus-operator/pkg/k8s/spec"
 	lc "github.com/piraeusdatastore/piraeus-operator/pkg/linstor/client"
 
@@ -50,9 +50,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-const linstorControllerFinalizer = "finalizer.linstor-controller.linbit.com"
+const (
+	linstorControllerFinalizer = "finalizer.linstor-controller.linbit.com"
 
-// var log = logf.Log.WithName("controller_linstorcontrollerset")
+	// requeue reconciliation after connectionRetrySeconds
+	connectionRetrySeconds = 10
+)
 
 func init() {
 	logrus.SetFormatter(&logrus.TextFormatter{})
@@ -112,18 +115,6 @@ type ReconcileLinstorControllerSet struct {
 	linstorClient *lc.HighLevelClient
 }
 
-func newCompoundErrorMsg(errs []error) []string {
-	errStrs := make([]string, 0)
-
-	for _, err := range errs {
-		if err != nil {
-			errStrs = append(errStrs, err.Error())
-		}
-	}
-
-	return errStrs
-}
-
 // Reconcile reads that state of the cluster for a LinstorControllerSet object and makes changes based
 // on the state read and what is in the LinstorControllerSet.Spec
 // TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
@@ -135,16 +126,19 @@ func newCompoundErrorMsg(errs []error) []string {
 // sense to put elsewhere, so don't lint it for cyclomatic complexity.
 // nolint:gocyclo
 func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := logrus.WithFields(logrus.Fields{
+	log := logrus.WithFields(logrus.Fields{
 		"resquestName":      request.Name,
 		"resquestNamespace": request.Namespace,
 	})
 
-	reqLogger.Info("CS Reconcile: Entering")
+	log.Info("CS Reconcile: Entering")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
 
 	// Fetch the LinstorControllerSet instance
 	pcs := &piraeusv1alpha1.LinstorControllerSet{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, pcs)
+	err := r.client.Get(ctx, request.NamespacedName, pcs)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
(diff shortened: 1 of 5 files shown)


pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -207,13 +189,13 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
 
 	markedForDeletion := pcs.GetDeletionTimestamp() != nil
 	if markedForDeletion {
-		finished, err := r.finalizeControllerSet(pcs)
+		finished, err := r.finalizeControllerSet(ctx, pcs)
```
JoelColledge, line 192: If we're going to go with this `TemporaryError` approach, we should adjust this method to use that too.

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -428,8 +393,24 @@ func (r *ReconcileLinstorControllerSet) reconcileControllerNodeWithControllers(p
...
 }
 
+func (r *ReconcileLinstorControllerSet) reconcileStatus(ctx context.Context, pcs *piraeusv1alpha1.LinstorControllerSet, errs []error) error {
+	log := logrus.WithFields(logrus.Fields{
+		"Name":      pcs.Name,
+		"Namespace": pcs.Namespace,
+	})
+	log.Debug("reconcile status")
+
+	log.Debug("reconcile error list")
```
JoelColledge, line 403: We don't really need 2 log messages directly one after the other. I would combine them.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -156,43 +158,14 @@ func (r *ReconcileLinstorNodeSet) Reconcile(request reconcile.Request) (reconcil
...
-	if pns.Spec.StoragePools == nil {
-		pns.Spec.StoragePools = &piraeusv1alpha1.StoragePools{}
-	}
-	if pns.Spec.StoragePools.LVMPools == nil {
-		pns.Spec.StoragePools.LVMPools = make([]*piraeusv1alpha1.StoragePoolLVM, 0)
-	}
-	if pns.Spec.StoragePools.LVMThinPools == nil {
-		pns.Spec.StoragePools.LVMThinPools = make([]*piraeusv1alpha1.StoragePoolLVMThin, 0)
-	}
+	log.Debug("fill in defaults in spec")
```
JoelColledge, line 168: This log message is kind of out of place. There are other messages when we actual perform the fill, so we can remove this.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -156,43 +158,14 @@ func (r *ReconcileLinstorNodeSet) Reconcile(request reconcile.Request) (reconcil
...
+	log.Debug("check if all required fields are filled")
 
-	logrus.WithFields(logrus.Fields{
-		"name":      pns.Name,
-		"namespace": pns.Namespace,
-		"spec":      fmt.Sprintf("%+v", pns.Spec),
-	}).Debug("NS Reconcile: found LinstorNodeSet")
-
-	if pns.Status.SatelliteStatuses == nil {
-		pns.Status.SatelliteStatuses = make([]*piraeusv1alpha1.SatelliteStatus, 0)
```
JoelColledge, line 184: Why can we remove these lines? We need to be robust against status serialization errors due to nil lists.

pkg/controller/linstornodeset/linstornodeset_controller.go
```diff
@@ -282,12 +256,27 @@ func (r *ReconcileLinstorNodeSet) Reconcile(request reconcile.Request) (reconcil
 	errs := r.reconcileAllNodesOnController(ctx, pns)
 
 	statusErr := r.reconcileStatus(ctx, pns, errs)
+	if statusErr != nil {
+		return reconcile.Result{}, statusErr
```
JoelColledge, line 260: It would make sense to have a similar warning here to the one in the controller set where we log the suppressed errors.

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -300,40 +270,20 @@ func (r *ReconcileLinstorControllerSet) Reconcile(request reconcile.Request) (re
...
-	if len(compoundErrorMsg) != 0 {
-		compoundError = fmt.Errorf("CS Reconcile: requeuing reconcile loop for the following reason(s): %s", strings.Join(compoundErrorMsg, " ;; "))
+	statusErr := r.reconcileStatus(ctx, pcs, errs)
+	if statusErr != nil {
+		log.Warnf("failed to update status. original errors: %v", errs)
+		return reconcile.Result{}, statusErr
 	}
 
-	return reconcile.Result{RequeueAfter: time.Minute * 1}, compoundError
+	return reconcileutil.ToReconcileResult(errs...)
```
JoelColledge, line 283: Maybe log something to indicate what we are returning here? I don't think we can see that otherwise if there are no actual errors returned. Same for node set controller.

pkg/controller/linstornodeset/linstornodeset_controller.go
WanzenBug: The only point where we should update the status field is in the `reconcileStatus` method. Otherwise there is no guarantee that the spec will be saved/updated.

I made some changes to `reconcileStatus()` that ensure that all lists are initialized as empty.
This should be enough, as an "empty" status will not be serialized at all, and if it is set, it will be by our `reconcileStatus()`

WanzenBug:
@JoelColledge please re-review.
I updated the finalizer logic a bit on both resources, so please review with extra care :)

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -428,58 +404,68 @@ func (r *ReconcileLinstorControllerSet) reconcileControllerNodeWithControllers(p
...
+	if finalizerErr != nil {
+		return reconcileutil.ToReconcileResult(finalizerErr)
+	}
+
+	err := r.deleteFinalizer(ctx, pcs)
+
+	return reconcile.Result{}, err
 }
 
 // mayFinalizeControllerSet returns whether the controller set may be finalized currently
```
JoelColledge, line 453: This comment is no longer valid

pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -428,58 +404,68 @@ func (r *ReconcileLinstorControllerSet) reconcileControllerNodeWithControllers(p
...
-		}
+	finalizerErr := r.mayFinalizeControllerSet(ctx, pcs)
 
-		// Remove finalizer. Once all finalizers have been
-		// removed, the object will be deleted.
-		log.Info("CS finalizing finished, removing finalizer")
-		if err := r.deleteFinalizer(pcs); err != nil {
-			return false, err
-		}
+	log.Info("CS finalizing finished, removing finalizer")
```
JoelColledge, line 436: Doesn't this belong with the `deleteFinalizer` line?

WanzenBug:
rebased to latest master

PR #40, opened 2020-06-17
helm: replace etcd chart

WanzenBug:
Use a custom version of the incubator/etcd chart.
Ongoing issues with bitnami/etcd [1] meant a node failure basically
permanently killed the etcd. The new chart taken from [2], updated
for our use-case, is able to recover from node failures

[1]: https://github.com/bitnami/charts/issues/1908
[2]: https://github.com/helm/charts/tree/master/incubator/etcd

### DIFF
--- a/charts/piraeus/charts/etcd/Chart.yaml
+++ b/charts/piraeus/charts/etcd/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+appVersion: 3.4.9
+description: Distributed reliable key-value store for the most critical data of a
+  distributed system.
+home: https://github.com/coreos/etcd
+icon: https://raw.githubusercontent.com/coreos/etcd/master/logos/etcd-horizontal-color.png
+maintainers:
+- email: lachlan@deis.com
+  name: lachie83
+name: etcd
+sources:
+- https://github.com/coreos/etcd
+version: 0.7.4
--- a/charts/piraeus/charts/etcd/README.md
+++ b/charts/piraeus/charts/etcd/README.md
@@ -1,0 +1,187 @@
+# Etcd Helm Chart
+
+Credit to https://github.com/ingvagabund. This is an implementation of that work
+
+* https://github.com/kubernetes/contrib/pull/1295
+
+## Prerequisites Details
+* Kubernetes 1.5 (for `StatefulSets` support)
+* PV support on the underlying infrastructure
+* ETCD version >= 3.0.0
+
+## StatefulSet Details
+* https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/
+
+## StatefulSet Caveats
+* https://kubernetes.io/docs/concepts/abstractions/controllers/statefulsets/#limitations
+
+## Chart Details
+This chart will do the following:
+
+* Implemented a dynamically scalable etcd cluster using Kubernetes StatefulSets
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```bash
+$ helm install --name my-release etcd
+```
+
+## Configuration
+
+The following table lists the configurable parameters of the etcd chart and their default values.
+
+| Parameter                           | Description                          | Default                                            |
+| ----------------------------------- | ------------------------------------ | -------------------------------------------------- |
+| `image.repository`                  | Container image repository           | `gcr.io/etcd-development/etcd`                     |
+| `image.tag`                         | Container image tag                  | `v3.4.9`                                           |
+| `image.pullPolicy`                  | Container pull policy                | `IfNotPresent`                                     |
+| `replicas`                          | k8s statefulset replicas             | `3`                                                |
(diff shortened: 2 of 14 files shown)


JoelColledge:
I'm getting an error when installing without persistence:
```
$ helm install piraeus-op ./charts/piraeus --set etcd.persistentVolume.enabled=false
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec.volumeClaimTemplates[0].metadata): unknown field "emptyDir" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
```

Do you see this too? I'm using Helm v3.2.4.

Using `helm template` shows that this is being generated, which fits the error:
```
  volumeClaimTemplates:
  - metadata:
      name: datadir
      emptyDir: {}
```

Looks like the `if` constructs are not quite in the right places.

WanzenBug:
Yeah, I guess I messed something up when updating the chart. I am still working on some improvements.

JoelColledge:
This etcd instance fails to start if there are fewer compute nodes available than replicas. That makes sense. Maybe we should set the default replica count to 1 to make the initial experience smooth for people who test on small clusters like Minikube?

WanzenBug:
Changed the default etcd cluster size to 1

charts/piraeus/charts/etcd/values.yaml
```diff
@@ -0,0 +1,78 @@
...
+# Declare name/value pairs to be passed into your templates.
+# name: value
+
+peerPort: 2380
+clientPort: 2379
+component: "etcd"
+replicas: 3
+image:
+  repository: "k8s.gcr.io/etcd-amd64"
+  tag: "3.2.26"
```
JoelColledge, line 12: This is quite an old release (11 Jan 2019). Have you looked for newer images?

charts/piraeus/charts/etcd/README.md
```diff
@@ -0,0 +1,188 @@
...
+This chart will do the following:
+
+* Implemented a dynamically scalable etcd cluster using Kubernetes StatefulSets
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```bash
+$ helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
```
JoelColledge, line 28: Doesn't make sense to refer to this URL.

charts/piraeus/charts/etcd/README.md
```diff
@@ -0,0 +1,188 @@
...
+| `image.tag`                         | Container image tag                  | `3.2.26`                                           |
+| `image.pullPolicy`                  | Container pull policy                | `IfNotPresent`                                     |
+| `replicas`                          | k8s statefulset replicas             | `3`                                                |
+| `resources`                         | container required resources         | `{}`                                               |
+| `clientPort`                        | k8s service port                     | `2379`                                             |
+| `peerPorts`                         | Container listening port             | `2380`                                             |
+| `storage`                           | Persistent volume size               | `1Gi`                                              |
+| `storageClass`                      | Persistent volume storage class      | `anything`                                         |
+| `affinity`                          | affinity settings for pod assignment | `{}`                                               |
```
JoelColledge, line 47: This default is incorrect. Not that important.

charts/piraeus/charts/etcd/values.yaml
WanzenBug: I only did a very brief search for newer tags on gcr.io, but have not found anything. I will try https://quay.io/repository/coreos/etcd

JoelColledge: That's an improvement. Anything "coreos" related might be abandoned or moved since the project in its previous form has been stopped. We could even continue to use the bitnami images, since the issues were in the chart.

WanzenBug: Actually, it seems the etcd team uses [two registries](https://github.com/etcd-io/etcd/releases/tag/v3.4.9):
* primary: https://gcr.io/etcd-development/etcd
* secondary: https://quay.io/coreos/etcd
So it looks like the coreos image won't go away anytime soon. I will still switch to the primary source

WanzenBug: Changed to `gcr.io/etcd-development/etcd:v3.4.9`

charts/piraeus/charts/etcd/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+appVersion: 3.2.26
```
JoelColledge, line 2: `appVersion` should match up with the image tag.

WanzenBug: good catch :+1:

charts/piraeus/charts/etcd/values.yaml
WanzenBug: from my experience the updated image seems to work

PR #49, opened 2020-07-06
use affinity and tolerations instead of node label

WanzenBug:
Affinity and Tolerations for all pods can now be configured via resources.
This changes the default behaviour of satellite pods. A guide how to restore
the old behaviour is included in the CHANGELOG.


WanzenBug:
For reference, the only "nice" solution I can think of is not an option: https://github.com/kubernetes/kubernetes/issues/62872
With `$ref` it would be possible to have validation without polluting our generated CRDs.

kermat:
@JoelColledge I think I like it. We're often labeling all nodes in the k8s cluster, so it's one less step there. In the scenario where a user has dedicated k8s nodes for different "workgroups" that share a control plane, and they only want to use LINSTOR in a specific "workgroup", they would already have some affinities/taints/tolerations in place. I think users who require this type of control would rather use their existing methods of constraint, while users without this requirement will continue to wonder why they have to label all their nodes.

PR #50, opened 2020-07-08
use deployment to configure linstor-controller

WanzenBug:
By using a deployment, a controller can be re-scheduled in case of a node
failure without human intervention. Since multiple LINSTOR controllers
cannot manage the same set of nodes at the same time, we cap the replica
count at 1 and use the "Recreate" deployment strategy.

In case of a node going suddenly offline, a new pod is scheduled after the
default timeout (300s). The old pod will still be marked as running, but it
won't receive any traffic (the node is offline, and the service endpoint would
not route traffic in any case).

The possibility of multiple pods associated with a controller deployment being
available in the Kubernetes API at the same time needed some rework in the
ControllerSet registration and status update code.

### DIFF
--- a/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
+++ b/pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
@@ -92,7 +92,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, &handler.EnqueueRequestForOwner{
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
 		IsController: true,
 		OwnerType:    &piraeusv1alpha1.LinstorControllerSet{},
 	})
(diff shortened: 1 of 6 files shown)


pkg/controller/linstorcontrollerset/linstorcontrollerset_controller.go
```diff
@@ -302,81 +302,132 @@ func (r *ReconcileLinstorControllerSet) reconcileControllers(ctx context.Context
...
+			pcs.Status.ControllerStatus.RegisteredOnController = true
+		}
+	}
+
+	log.Debug("fetch information about storage nodes")
 
 	nodes, err := r.linstorClient.GetAllStorageNodes(ctx)
 	if err != nil {
-		return fmt.Errorf("CS unable to get cluster storage nodes: %v", err)
+		log.Warnf("unable to get LINSTOR storage nodes: %v", err)
```
JoelColledge, line 430: If `err != nil`, I wouldn't trust the other values returned. Maybe set `nodes` to an empty slice explicitly here?

WanzenBug: nice catch!

PR #51, opened 2020-07-09
customizable image registry and pull policy

alexzhc:
Proposed changes:
```
1. make images customizable: csi-snapshotter-controller, kube-scheduler-amd64
2. make linstor-controller pod TerminationGracePeriodSeconds=0, so that it can be automatically rescheduled after node goes down
3. make imagePullPolicy customizable
4. created a value.cn.yaml for chinese users who cannot stably access quay.io, gcr.io and docker hub.
```

TO-DOs:
```
1. add resource limit for each Pod, as some k8s platform enforce quota so that pods without specified resource limit cannot come up
2. the csi-controller pod should be deployed by statefulset with leader election, instead of a single-pod deployment
3. there should a quick start in the readme which gives less options. 
4. Regarding the failover of linstor-controller pod, even with TerminationGracePeriodSeconds=0, it will take about 5 minutes before kubernetes reschedules the pod, which is too slow for a storage system. I think we should explore use Kubernetes built-in leader-election for its HA.ref: https://tunein.engineering/implementing-leader-election-for-kubernetes-pods-2477deef8f13
```

### DIFF
--- a/pkg/apis/piraeus/v1alpha1/linstornodeset_types.go
+++ b/pkg/apis/piraeus/v1alpha1/linstornodeset_types.go
@@ -18,7 +18,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -59,6 +59,10 @@ type LinstorNodeSetSpec struct {
 	// drbdRepoCred is the name of the kubernetes secret that holds the credential for the DRBD repositories
 	DrbdRepoCred string `json:"drbdRepoCred"`
 
+	// Pull policy applied to all pods started from this controller
+	// +optional
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy"`
+
 	// satelliteImage is the image (location + tag) for the LINSTOR satellite container
 	SatelliteImage string `json:"satelliteImage"`
 
@@ -73,12 +77,12 @@ type LinstorNodeSetSpec struct {
 	// Affinity for scheduling the satellite pods
 	// +optional
 	// +nullable
-	Affinity *v1.Affinity `json:"affinity"`
+	Affinity *corev1.Affinity `json:"affinity"`
 
 	// Tolerations for scheduling the satellite pods
 	// +optional
 	// +nullable
-	Tolerations []v1.Toleration `json:"tolerations"`
+	Tolerations []corev1.Toleration `json:"tolerations"`
 
 	LinstorClientConfig `json:",inline"`
 }
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,3 +36,9 @@ repos:
         files: '^pkg/apis/.*\.go$'
         pass_filenames: false
         entry: make crds
+      - id: generate-cn-values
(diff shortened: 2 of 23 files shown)


charts/piraeus/values.yaml
```diff
@@ -1,38 +1,52 @@
...
+    
 stork:
   enabled: true
-  image: docker.io/linbit/stork:latest
+  storkImage: "docker.io/linbit/stork:latest"
+  schedulerImage: "gcr.io/google_containers/kube-scheduler-amd64"
   replicas: 1
 csi-snapshotter:
   enabled: true # <- enable to add k8s snapshotting CRDs and controller. Needed for CSI snapshotting
+  image: "daocloud.io/piraeus/snapshot-controller:v2.0.1"
```
WanzenBug, line 21: should be `quay.io/k8scsi/snapshot-controller:v2.1.0`?

pkg/apis/piraeus/v1alpha1/linstornodeset_types.go
```diff
@@ -60,9 +61,11 @@ type LinstorNodeSetSpec struct {
 
 	// satelliteImage is the image (location + tag) for the LINSTOR satellite container
 	SatelliteImage string `json:"satelliteImage"`
+	SatelliteImagePullPolicy corev1.PullPolicy `json:"SatelliteImagePullPolicy"`
```
WanzenBug, line 64: json tag should be lower case:
```
`json:"satelliteImagePullPolicy"`
```

pkg/apis/piraeus/v1alpha1/linstornodeset_types.go
```diff
@@ -60,9 +61,11 @@ type LinstorNodeSetSpec struct {
...
 
 	// kernelModImage is the image (location + tag) for the LINSTOR/DRBD kernel module injector container
 	KernelModImage string `json:"kernelModImage"`
+	KernelModImagePullPolicy corev1.PullPolicy `json:"KernelModImagePullPolicy"`
```
WanzenBug, line 68: json tag should be lower case:
```
`json:"kernelModImagePullPolicy"`
```

WanzenBug:
Also there is a merge conflict in `charts/piraeus/values.yaml`, please resolve that. It should just be a few lines added to our master since you forked the repository

alexzhc:
@WanzenBug It might be better to just implement a **global ImagePullPolicy** for all the containers, instead of tuning each one. How do you think?

WanzenBug:
That's probably the way to go. The main use case I see is setting `imagePullPolicy: never`, in which case you want it enabled for everything. For the other cases k8s has sensible defaults depending on the user used `:latest` or not

WanzenBug:
Hey @alexzhc 
I worked some more on this:
* I made the pull policy optional in the CRDs. That means old resources should continue to work
* I updated our Makefile/commit hooks to keep `values.cn.yaml` in sync with `values.yaml`
* Updates to README and CHANGELOG

Could you please make sure that the referenced images are available on daocloud.io?

I'll merge this tomorrow if you don't have anything to add

alexzhc:
sure, I would suggest a global `imagePullPolicy: IfNotPresent` for users. Dev can change it to `Always` for the testing.

alexzhc:
Please update the image versions both in value.yaml and value.cn.yaml as you see fit. I will make sure push to the local registry here.

WanzenBug:
Thanks!

Issue #52, opened 2020-07-09
Two controllers?


kvaps:
Hi, I just tried new piraeus-operator, and found that it is deployed two controllers, one as `kind: Deployment` another one as `kind: StatefulSet`:

```bash
# k get deploy
NAME                         READY   UP-TO-DATE   AVAILABLE   AGE
piraeus-op-cs-controller     1/1     1            1           12h
piraeus-op-csi-controller    1/1     1            1           10d
piraeus-op-operator          1/1     1            1           10d
piraeus-op-stork             1/1     1            1           10d
piraeus-op-stork-scheduler   1/1     1            1           10d

# k get sts
NAME                       READY   AGE
piraeus-op-cs-controller   0/1     10d
piraeus-op-etcd            1/1     10d
snapshot-controller        1/1     10d

# k get pod -l app=piraeus-op-cs,role=piraeus-controller
NAME                                        READY   STATUS    RESTARTS   AGE
piraeus-op-cs-controller-0                  1/1     Running   0          5m25s
piraeus-op-cs-controller-75fc99d964-c5z87   1/1     Running   0          12h
```

```bash
root@piraeus-op-cs-controller-0:/# linstor n l
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
┊ Node                                      ┊ NodeType   ┊ Addresses                 ┊ State                     ┊
╞════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
┊ ks1                                       ┊ SATELLITE  ┊ 172.16.0.6:3366 (PLAIN)   ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ ks2                                       ┊ SATELLITE  ┊ 172.16.0.7:3366 (PLAIN)   ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ ks3                                       ┊ SATELLITE  ┊ 172.16.0.8:3366 (PLAIN)   ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ ks4                                       ┊ SATELLITE  ┊ 172.16.0.9:3366 (PLAIN)   ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ piraeus-op-cs-controller-0                ┊ CONTROLLER ┊ 10.244.3.216:3366 (PLAIN) ┊ OFFLINE                   ┊
[...]

PR #54, opened 2020-07-13
add release preparatation target

WanzenBug:
Adds a new `make` target that:
* Creates a new commit with
  * increased chart versions
  * matching operator image version
  * updates changelog by replacing `Unreleased` with the new version
* Creates another commit, which:
  * adds the unreleased section to the changelog
  * Sets the operator image version to the dev setting

This helps address the issue brought up in #52. For any release, the images are all pinned to a specific tag. For day to day development, we can still use the `:latest` image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,3 +31,45 @@ deep-copy:
 crds:
 	operator-sdk generate crds
 	mv ./deploy/crds/* ./charts/piraeus/crds
+
+release:
+	# check that VERSION is set
+	@if [ -z "$(VERSION)" ]; then \
+		echo "Usage: make prepare-release VERSION=<version>" >&2 ; \
+		exit 1 ; \
+	fi
+	# check that version has expected format
+	@if ! echo -e "$(VERSION)" | grep -qP '^\d+\.\d+\.\d+$$' ; then \
+		echo "version format: <maj>.<min>.<patch>" >&2 ; \
+		exit 1 ; \
+	fi
+	# check that version does not exist
+	@if git rev-parse "v$(VERSION)" >/dev/null 2>&1 ; then \
+		echo "version v$(VERSION) already exists" >&2 ; \
+		exit 1 ; \
+	fi
+	# check that working tree is clean
+	@if ! git diff-index --quiet HEAD -- ; then \
+		echo "Refusing to create release from dirty repository" >&2 ; \
+		exit 1; \
+	fi
+	# replace changelog header "Unreleased" with version and replace link target
+	sed 's/^## \[Unreleased\]/## [v$(VERSION)] - $(shell date +%Y-%m-%d)/' -i CHANGELOG.md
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,3 +31,45 @@ deep-copy:
...
 	mv ./deploy/crds/* ./charts/piraeus/crds
+
+release:
+	# check that VERSION is set
+	@if [ -z "$(VERSION)" ]; then \
+		echo "Usage: make prepare-release VERSION=<version>" >&2 ; \
+		exit 1 ; \
+	fi
+	# check that version has expected format
+	@if ! echo -e "$(VERSION)" | grep -cP '^\d+\.\d+\.\d+$$' > /dev/null ; then \
```
rck, line 42: using the count + /dev/null feels a bit uncommon. What abuot dropping `-c` and using `-q`?

Makefile
```diff
@@ -31,3 +31,45 @@ deep-copy:
...
+		exit 1 ; \
+	fi
+	# check that version does not exist
+	@if git rev-parse "v$(VERSION)" >/dev/null 2>&1 ; then \
+		echo "version v$(VERSION) already exists" >&2 ; \
+		exit 1 ; \
+	fi
+	# check that working tree is clean
+	@if ! git diff-index --quiet HEAD -- ; then \
+			echo "Refusing to create release from dirty repository" >&2 ; \
```
rck, line 53: does not really matter, but compared to the others this has one additional indent level.

Makefile
```diff
@@ -31,3 +31,45 @@ deep-copy:
...
+	yq w -i charts/piraeus/Chart.yaml version $(VERSION)
+	yq w -i charts/piraeus/Chart.yaml appVersion $(VERSION)
+	# set operator image to tagged version
+	yq w -i charts/piraeus/values.yaml operator.image "quay.io/piraeusdatastore/piraeus-operator:$(VERSION)"
+	# commit as current release + tag
+	git commit -aevm "Release v$(VERSION)"
+	git tag v$(VERSION)
+	# add "Unreleased" section at top + create comparison link against current master
+	sed 's/^## \[v$(VERSION)\]/## [Unreleased]\n\n## [v$(VERSION)]/' -i CHANGELOG.md
+	echo "[Unreleased]: https://github.com/piraeusdatastore/piraeus-operator/compare/v$(VERSION)...HEAD" >> CHANGELOG.md
```
rck, line 71: maybe lost in magic, but doesn't that grow at the tail over time? I see how you handle the top with the seds, but this is always added at the bottom, isn't it?

Makefile
WanzenBug: TIL `-q` :sweat_smile:

Makefile
WanzenBug: Yeah, it does. However I think it does not really matter:
* its only the link target, it will never be display in the rendered markdown
* its much easier to add it at the end then to find/replace the correct section with sed

I guess in the grand scheme of things there are already so many regexp in this section that one more wouldn't really matter. so...your choice

Issue #56, opened 2020-07-14
Enable linstor-controller to run with multiple replicas


WanzenBug:
For HA we want to have multiple instances of a LINSTOR controller running, with internal leader election to determine the active one.

This is part of #55.
See also the discussion around #50, #52, #53

dimm0:
Is this supported? There's no `replicas` field in the linstorcluster spec.controller CRD

WanzenBug:
No, it was always a hack. There can only be one controller running at a time. The way we achieve faster failover is using tolerations that cause the Pod to be quickly rescheduled should the node experience issues.

dimm0:
It takes a couple minutes for my pod to start.. we have more than 1000 volumes in the cluster.

But ok, thanks for the info

PR #62, opened 2020-07-21
crds: s/controllerset/controller/ + s/nodeset/satelliteset/

### LINKED ISSUE #59: Syncronize names with LINSTOR
# What
Before releasing v1.0, we want to synchronize names between the operator and LINSTOR.
The current proposal is to rename these CRDs:
* `LinstorNodeSet` -> `LinstorSatelliteSet`
* `LinstorControllerSet` -> `LinstorController`

# Why
Some names used by the operator obscure the actual use. LinstorNodeSet should actually be LinstorSatelliteSet. LinstorControllerSet is only ever 1 (active) controller, so it is not a set

# For whom
* For our users, to get coherent documentation between LINSTOR and the operator
* For us, to have similar concepts named similarly

# Done when:
* [ ] The renamed CRDs are used by the operator
* [ ] There is documentation on the renaming
* [ ] There is an automatic upgrade path

WanzenBug:
Rename 2 CRDS:
* LinstorControllerSet -> LinstorController
* LinstorNodeSet -> LinstorSatelliteSet

Regarding LinstorController: the change was made to indicate
that this resource controls a single controller (which may have multiple
replicas, but still). Old resources will be automatically converted into
the newer form.

Regarding LinstorSatelliteSet: the change was made to line up with other
LINSTOR resources. While for many cases one satellite will run per node,
this will not always be the case. Old resource will be automatically converted
into the newer form.

Fixes #59


JoelColledge:
How does this relate to #56? Maybe we will soon have multiple controller pods at the Kubernetes level :upside_down_face: I guess LinstorController could still be considered a better name, because it is just one service.

What is the plan for removing the old CRDs and legacy controllers?

WanzenBug:
> How does this relate to #56? Maybe we will soon have multiple controller pods at the Kubernetes level. I guess LinstorController could still be considered a better name, because it is just one service.

I'd argue exactly like that: from a high-level perspective, it's still just a single controller (with backups in case of failure). From an outsiders perspective this is an implementation detail, they only ever talk to "the" controller

> What is the plan for removing the old CRDs and legacy controllers?

I'd keep the legacy controller for the 1.0 release and remove it on the next minor release. That way we can create update instructions like "if < 1.0, upgrade to 1.0 first".

PR #69, opened 2020-07-27
Upgrade CRDs to v1

WanzenBug:
Upgrading the CRDs to v1 marks them as stable, which is a requirement
for the v1.0 release of the operator.

The actual types are simple copies of the current v1alpha1 versions.
They are marked as "storageversions" i.e. kubernetes will store all
resources as v1.

The operator itself already works on v1 resources, except for the
legacy controller (which will write v1alpha versions that are directly
converted to v1 by kubernetes).

To ensure that types are compatible, a simple test serializes
v1alpha types and deserializes them as v1.


WanzenBug:
re: changes to CRDs between RC and v1... I would decide on that per case. If its not a feature/fix that we **need** in our v1, I wouldn't change the CRD. If its broken and needs repair, then obviously we should do it before v1.

PR #73, opened 2020-08-04
support multiple replicas for linstor-controller

WanzenBug:
Running multiple replicas requires special support from the linstor
controller. The controller container will start a leader election
process when it detects the presence of the `K8S_AWAIT_ELECTION_*`
variables.

The election process determines which pod is allowed to start the
linstor-controller process. Only this pod will be marked as ready
and will receive traffic from the k8s service object.

Should the leader crash or the node its running on goes offline,
a new leader will be elected and allowed to start the controller process.
Note: in case the full node goes offline, the old pod will still be marked
as ready. By using `ClusterIP: ""` on our service, we ensure we create an actual
proxy (which automatically chooses a responding pod) instead of each client
having to deal with multiple DNS responses which may or may not respond.

See also:
* https://github.com/piraeusdatastore/piraeus/pull/42
* https://github.com/LINBIT/k8s-await-election/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking
+
+* The LINSTOR controller image given in `operator.controller.controllerImage` has to have
+  its entrypoint set to [`k8s-await-election v0.2.0`](https://github.com/LINBIT/k8s-await-election/)
+  or newer. Learn more in the [upgrade guide](./UPGRADE.md#upgrade-from-v10-to-head).
+
 ### Added
 
+* LINSTOR controller can be started with multiple replicas. See [`operator.controller.replicas`](./doc/helm-values.adoc#operatorcontrollerreplicas).
+  NOTE: This requires support from the container. You need `piraeus-server:v1.8.0` or newer.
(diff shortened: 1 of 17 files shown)


JoelColledge:
We can get this merged now that we are no longer in an RC phase. Rebase needed, and the Helm keys need to be added to the new docs.

WanzenBug:
I need to check if environment variables are passed to the ExecProbe. Then we could maybe find a cleaner solution

WanzenBug:
So environment variables are available in ExecProbe

During testing I noticed a few additional issues:
* The controller could not handle multiple running controllers at once. I just removed that check, as in this new version multiple controllers are expected.
* Rolling upgrades, are not really working. Probably because the new container won't be `Ready` until it is elected, but the old leader is still running, so the upgrade process basically gets stuck.

The second issue can be traced to the way we (ab-)use the `Ready`mechanism. Not sure how we should deal with this yet

JoelColledge (approved):
Looks good. Just waiting for internal tests to succeed.

WanzenBug:
I want to polish our documentation a bit

WanzenBug:
@JoelColledge I think this is finally ready

JoelColledge (approved):
Looks good to me. Just waiting for internal tests to pass before merging.

JoelColledge:
:tada:

PR #74, opened 2020-08-05
Create a reference table for all helm settings

WanzenBug:
The table can be used to quickly look up
* default values
* allowed values
* a description of what the value does.

It is meant as a complementary resource to the other documentation, which
focuses on covering every possible setting in a easy to navigate way.

### DIFF
--- a/doc/helm-values.adoc
+++ b/doc/helm-values.adoc
@@ -1,0 +1,348 @@
+= Helm Reference Table
+
+== Global Settings
+
+=== `drbdRepoCred`
+
+Default:: `drbdiocred`
+Valid values:: secret name
+Description:: Names a secret containing registry credentials to pull LINSTOR container images.
+
+=== `global.imagePullPolicy`
+Default:: `IfNotPresent`
+Valid values::
+* `""`
+* `Always`
+* `IfNotPresent`
+* `Never`
+Description:: Global pull policy to apply to all images. Can be set to `""` to use kubernetes default behaviour. See https://kubernetes.io/docs/concepts/containers/images/#updating-images[pull policy].
+
+=== `linstorHttpsClientSecret`
+Default:: `""`
+Valid values:: secret name
+Description:: References the secret to use when configuring LINSTOR clients to use HTTPS. Check out link:./security.md#configuring-secure-communications-for-the-linstor-api[the security guide]
+
+=== `linstorHttpsControllerSecret`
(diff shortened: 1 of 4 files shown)


doc/helm-values.adoc
```diff
@@ -0,0 +1,381 @@
...
+[cols="10,10,20a,50a"]
+|===
+|setting |default value |valid values |description
+
+|`image`
+|`quay.io/piraeusdatastore/piraeus-operator:<tag>`
+|image ref
+|Image to use for the operator deployment.
+
+NOTE: prefer to use a pinned version (i.e. `:v1.0.0`) over a rolling tag (i.e. `:latest`). During development, the default value will be `:latest`, on release it will match the release tag.
```
JoelColledge, line 47: Some sections don't wrap properly when rendered by GitHub. Presumably due to these `NOTE`s.

doc/helm-values.adoc
```diff
@@ -0,0 +1,381 @@
...
+|`priorityClassName` |`""` |https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass[priority class] name |Name of the priority class with which the LINSTOR components should be scheduled.
+|===
+
+== Operator (`operator` section)
+
+[cols="10,10,20a,50a"]
+|===
+|setting |default value |valid values |description
+
+|`image`
```
JoelColledge, line 42: At https://github.com/bitnami/charts/tree/master/bitnami/etcd they have the whole key which I found useful because then you can just copy it into a `--set` flag. That is less useful if you are writing a yaml values file, but I find it overall more explicit and easier to understand at a glance. What do you think?

doc/helm-values.adoc
```diff
@@ -0,0 +1,381 @@
...
+|Affinity settings for satellite pods. Can be used to restrict the pods to specific nodes.
+
+|`automaticStorageType`
+| `None`
+|
+* `None`
+* `LVM`
+* `LVMTHIN`
+* `ZFS`
+| Automatically create a storage pool of the specified type. Check the link:./storage.md#preparing-physical-devices[storage guide].
```
JoelColledge, line 276: `Automatically create storage pools`

doc/helm-values.adoc
WanzenBug: It started out with the full key, but github's rendering is even worse in that case

PR #79, opened 2020-08-06
automatically create yaml deployment files from helm

WanzenBug:
Add a new make target that uses `helm template` to create a static
yaml deployment from the current helm chart. A pre-commit hook
makes sure that the files stay up-to-date with chart development.


deploy/piraeus/charts/etcd/templates/service.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Source: piraeus/charts/etcd/templates/service.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
+metadata:
+  name: piraeus-op-etcd
+  labels:
+    heritage: "Helm"
```
JoelColledge, line 11: I guess this means that we could manage this deployment with Helm :smile: But that's not a problem.

JoelColledge:
Rebase needed.

PR #81, opened 2020-08-06
fixup for help output from legacy-check.yaml

kermat:
The new name for `operator.satelliteSet.kernelModImage` is
`operator.satelliteSet.kernelModuleInjectionImage`, but the help output
suggests using the old name.


WanzenBug:
Thanks for spotting this!

PR #89, opened 2020-09-03
remove legacy crds

WanzenBug:
This commit drops support for legacy CRDs "LinstorControllerSet" and
"LinstorNodeSet". They were deprecated before v1.0.0, which will be
the last version that supports them.

The pull secret can be made optional, as the default images are available
without credentials.

This also means we can drop some custom validation done in the operator
code. This can be handled by the OpenAPI schema.


JoelColledge (approved):
Looks good to me. Just waiting for internal tests before merging.

deploy/piraeus/templates/operator-deployment.yaml
```diff
@@ -39,4 +39,4 @@ spec:
           resources:
             {}
       imagePullSecrets:
-      - name: drbdiocred
+      - name:
```
WanzenBug, line 42: Need to clean this up: this should not be empty

WanzenBug:
@JoelColledge ready

JoelColledge (approved):

PR #93, opened 2020-09-21
Fix etcd volume for non-root etcd containers

WanzenBug:
Kubernetes volumes are mounted for uid 0 by default. This meant
that non-root containers could not write to these volumes. Our etcd
deployment needs to persist data, so running etcd as non-root user
would not work.

This commit introduces two separate fixes for this issues:

1. When using the default security context, fsGroup is set to 1000.
   Newly provisioned volumes will be "chgrp -R"-ed to that gid. This
   fixes above issue in case the storage provider supports it.

2. When using our pv-hostpath chart, 1) would not work, as hostpath
   volumes _always_ keep their host permissions. To make this work,
   a new helm hook is added that runs a job on volume creation,
   manually setting the host volumes permission.

### DIFF
--- a/charts/pv-hostpath/templates/pv.yaml
+++ b/charts/pv-hostpath/templates/pv.yaml
@@ -11,7 +11,8 @@ spec:
   accessModes:
   - ReadWriteOnce
   hostPath:
-    path: {{ required "A valid .Values.size entry required!" $.Values.path }}
+    path: {{ required "A valid .Values.path entry required!" $.Values.path }}
+    type: DirectoryOrCreate
   nodeAffinity:
     required:
       nodeSelectorTerms:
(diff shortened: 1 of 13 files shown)


charts/pv-hostpath/templates/pv.yaml
```diff
@@ -21,4 +22,41 @@ spec:
...
+            - {{ $.Values.uid | default 1000 }}:{{ $.Values.uid | default 1000 }}
+            - /volume
+          volumeMounts:
+            - mountPath: /volume
+              name: pv
+      volumes:
+        - name: pv
+          hostPath:
+            type: DirectoryOrCreate
+            path: {{ required "A valid .Values.size entry required!" $.Values.path }}
```
JoelColledge, line 60: Should be `.Values.path` in error message.

JoelColledge (commented):
Looks like a neat solution.

WanzenBug:
@kermat could you please give this a try?

kermat (approved):
This works for me. Tested with the default security context, and without (`--set global.setSecurityContext=false`); both work as expected now. Thanks @WanzenBug!

JoelColledge (approved):

PR #95, opened 2020-10-01
Update stork image selection to upstream v2.5.0

WanzenBug:


charts/piraeus/values.cn.yaml
```diff
@@ -18,7 +18,7 @@ csi-snapshotter:
   resources: {} # resource requirements for the cluster snapshot controller.
 stork:
   enabled: true
-  storkImage: daocloud.io/piraeus/stork:latest
+  storkImage: daocloud.io/piraeus/stork:2.5.0
```
JoelColledge, line 21: This doesn't exist yet. Should we just leave it as it is?

WanzenBug: The file is automatically generated by a commit hook. I would just keep the `2.5.0` and make sure the image is available on the next release.

JoelColledge (approved):

PR #100, opened 2020-10-13
Prepare v1.1.0

WanzenBug:
This include two small fixes:
* LinstorController and LinstorSatelliteSet now both take a serviceAccountName
      parameter, that is filled with a default value if left empty
* Overriding via environment variables now only updates empty values
      in the spec.

### DIFF
--- a/pkg/k8s/reconcileutil/update_from_env.go
+++ b/pkg/k8s/reconcileutil/update_from_env.go
@@ -18,11 +18,17 @@ func UpdateFromEnv(ctx context.Context, client client.Client, obj runtime.Object
 	changed := false
 
 	for i := range specs {
-		val, ok := os.LookupEnv(specs[i].Env)
-		if ok {
-			changed = true
-			*specs[i].Target = val
+		if *specs[i].Target != "" {
+			continue
 		}
+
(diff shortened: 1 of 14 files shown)


pkg/k8s/reconcileutil/update_from_env.go
```diff
@@ -19,10 +19,12 @@ func UpdateFromEnv(ctx context.Context, client client.Client, obj runtime.Object
 
 	for i := range specs {
 		val, ok := os.LookupEnv(specs[i].Env)
-		if ok {
-			changed = true
-			*specs[i].Target = val
+		if !ok || *specs[i].Target == "" {
```
JoelColledge, line 22: > Overriding via environment variables now only updates empty values in the spec.

It looks to me like this does the opposite. It skips the update if the value is empty.

Also, might be clearer if this condition is placed before `LookupEnv` since it is independent of that.

JoelColledge (approved):

pkg/k8s/reconcileutil/update_from_env.go
WanzenBug: Thanks for spotting this! That's what I get for not testing :sweat_smile:

PR #102, opened 2020-10-15
pv-hostpath: ensure permissions are applied on all nodes

WanzenBug:
This commit adds an affinity spec to jobs used to set the permissions
on created hostpath volumes. Without this, some nodes may not get their
permissions set for the created volume.


JoelColledge (dismissed):

WanzenBug:
Fixed the changelog entry, should be under "UNRELEASED"

JoelColledge:
Commit message should also be "With**out** this", if we're being picky.

JoelColledge (approved):

PR #105, opened 2020-10-23
Automatically prepare devices via storagePools configuration

WanzenBug:
This commit improves upon the automaticStorageType functionality
by integrating it with the existing storagePools config. Storage pools
can now directly specify which device they want to consume. This enables us to:
* Only using explicitly configured block devices
* Creating different storagepool types for different devices
* Setting extra parameters per storage pool
* Restore storage pools should the devices be prepared but LINSTOR lost its
  state (mainly useful for testing)

The use of the old automaticStorageType is deprecated.


doc/storage.md
```diff
@@ -13,9 +13,52 @@ are block devices that:
...
+    lvmThinPools:
+    - name: lvm-thin
+      thinVolume: thinpool
+      volumeGroup: linstor_thinpool
+      devicePaths:
+      - /dev/vdc
+      - /dev/vdd
+```
+
+Currently, this method support creation of LVM and LVMTHIN storage pools.
```
JoelColledge, line 34: `supports`

pkg/apis/piraeus/shared/linstor_types.go
```diff
@@ -78,47 +78,138 @@ type StoragePools struct {
...
 // LinstorSatelliteSet
 type StoragePoolLVM struct {
-	// Name of the storage pool.
-	Name string `json:"name"`
+	StoragePoolCommon `json:",inline"`
+
+	// Name of underlying lvm group
+	VolumeGroup string `json:"volumeGroup"`
+
+	// Enable VDO on the volume group
```
JoelColledge, line 118: Perhaps expand the `VDO` acronym here (the first place we use it)?

doc/storage.md
```diff
@@ -13,9 +13,52 @@ are block devices that:
 * do not contain partition information
 * have more than 1 GiB
 
-Any such device will be prepared according to the value of `operator.satelliteSet.automaticStorageType`. Devices are
-added to a storage pool based on the device name (i.e. all `/dev/nvme1` devices will be part of the pool
-`nvme1`)
+To enable automatic configuration of devices, set the `devicePaths` key on `storagePools` entries:
```
JoelColledge, line 16: It may now make more sense to place this section under the "Configuring storage pool creation" section because this builds on that concept.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -158,7 +161,8 @@ func (r *ReconcileLinstorSatelliteSet) Reconcile(request reconcile.Request) (rec
 
 	r.linstorClient, err = lc.NewHighLevelLinstorClientFromConfig(pns.Spec.ControllerEndpoint, &pns.Spec.LinstorClientConfig, getSecret)
 	if err != nil {
-		return reconcile.Result{}, err
```
JoelColledge, line 161: This will cause issues when there is an error because `r.linstorClient` will be `nil` but it is used later on.

doc/storage.md
JoelColledge, line 34: Perhaps we should add ZFS support at the same time. Otherwise we are deprecating the method to create ZFS pools without replacing it.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -457,19 +487,48 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAutomaticDeviceSetup(ctx context
...
+	if nodeSet.Spec.AutomaticStorageType == automaticStorageTypeNone {
+		return nil
+	}
 
+	for emptyDevice := range emptyDevices {
 		// Note: not found returns -1, so in this case name == path, which is exactly what we want
-		idx := strings.LastIndex(path, "/")
-		name := path[idx+1:]
+		idx := strings.LastIndex(emptyDevice, "/")
+		name := "dev-" + emptyDevice[idx+1:]
```
JoelColledge, line 528: I would find it strange to have a VG named `dev-sda`, but I'm not sure of a better alternative :upside_down_face: 

`pool-sda`? `vg-sda` (doesn't work for ZFS)?

doc/storage.md
WanzenBug: I actually started implementing it, but I'm pretty sure it never worked in the first place. There is some interaction between the ZFS kmod and the tools that does not seem to work in containers when I tried.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
WanzenBug: `autopool-sda`?

doc/storage.md
JoelColledge: Fair enough. If it never worked, no harm in deprecating it for now.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
JoelColledge: > `autopool-sda`?

Fine by me.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
WanzenBug: Good catch! I'll just update the status right there.

WanzenBug:
I'll take some more time to refine this PR (any maybe add some basic ZFS support)

WanzenBug:
@JoelColledge could you take another look. I added basic support for ZFS storage pools.

Sadly creation of zpools using LINSTOR does not work in containers for the time beeing, so I excluded the `devicePaths` section for ZFS.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -178,22 +196,24 @@ func (r *ReconcileLinstorSatelliteSet) Reconcile(request reconcile.Request) (rec
...
 	if err := r.addFinalizer(ctx, pns); err != nil {
-		return reconcile.Result{}, err
+		log.WithField("err", err).Warn("failed to add finalizer to resource")
+		errs = append(errs, err)
 	}
 
 	// Create the satellite configuration
 	configMap, err := reconcileSatelliteConfiguration(ctx, pns, r)
 	if err != nil {
-		return reconcile.Result{}, err
```
JoelColledge, line 187: I looks like a a nil `configMap` will also cause problems. Generally I'm a bit uncomfortable with this continue on errors approach. Even functions that don't return a value might have side-effects which are necessary for later steps to work. Could we restructure this function like this?

```
Get(pns)
err := doStuff(pns)
reconcileStatus(pns, err)
```

So that `doStuff` can return at any point and we still update the status?

If we really do perform independent steps here, then we could also build that into the code structure somehow to make it clearer what requires what.

doc/storage.md
```diff
@@ -74,10 +53,78 @@ helm install -f <file> charts/piraeus-op
...
+* `raidLevel` LVM raid level. Optional
+* `vdo` Enable [VDO] (requires VDO tools in the satellite). Optional
+* `vdoLogicalSizeKib` Size of the created VG (expected to be bigger than the backing devices by using VDO). Optional
+* `vdoSlabSizeKib` Slab size for VDO. Optional
+
+[VDO]: https://www.redhat.com/en/blog/look-vdo-new-linux-compression-layer
+
+#### `lvmThinPools` configuration
+* `name` name of the LINSTOR storage pool. Required
+* `volumeGroup` VG to use for the thin pool. Note: if `devicePaths` is set, this will be set to `linstor_$THINVOLUME`, to stay compatible with LINSTOR created pools. Required
```
JoelColledge, line 107: It is a bit strange for this to be required but in certain circumstances to not have a choice about the value. Would it be possible to make it optional, defaulting to the `linstor_$THINVOLUME` value? Maybe we should complain if neither `volumeGroup` nor `devicePaths` is set.

WanzenBug: Sadly the volumeGroup parameter isn't new, so marking it as optional would be considered a breaking change for K8s APIs.

What we can do is require a kind of XOR configuration, so you either have to set `volumeGroup: ""` and use `devicePaths` _or_ you set the volumeGroup, but then you can't use devicePaths

JoelColledge: I don't think we should reject storage pools where `devicePaths` is used and `volumeGroup` is equal to `linstor_$THINVOLUME`. That's what you have on a second reconciliation run, right? When `devicePaths` is used, we could reject values for `volumeGroup` that are not `""` or `linstor_$THINVOLUME`, and document here that `""` is OK. Does that work?

WanzenBug: sounds like a plan!

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
WanzenBug: So I reworked the SatelliteSet controller to follow this approach. 

Apart from cosmetic changes (renamed every satelliteset instance to `satelliteSet` instead of `nodeSet`, `pns`, etc...)
I also removed the linstor client from the reconciler object. This makes it easy to ensure that its initialized at usage site.

JoelColledge:
@WanzenBug I've lost track of the status of this. Is it ready for reviewing again and hopefully merging?

WanzenBug:
Yeah it's ready again

JoelColledge (approved):

PR #111, opened 2020-11-04
improve etcd deployment

WanzenBug:
* allow manual override of securityContext per component
  - etcd
  - csi-controller
  - stork
  - operator

* pv-hostpath integrated with PSP and SCC restrictions
* pv-hostpath prepares volumes for eventual fsGroup use + SELinux labels
* pv-hostpath jobs are able to run on master nodes by default
* etcd pods are able to run on master nodes by default


WanzenBug:
still missing some docs. Functionality should all be there already. I was able to use our chart to stand up an etcd instance on openshift

JoelColledge:
Great stuff.

WanzenBug:
Added docs on affinity + tolerations + added them when they were missing from the chart.

Also added an example configuration for openshift, which I successfully used to create a LINSTOR cluster on openshift

JoelColledge (approved):
Wow, never imagined that `pv-hostpath` would get so complicated.

PR #112, opened 2020-11-05
added openshift support

dkhachyan:


WanzenBug:
Hey! Thank you for your PR!

We are actually currently also in the process of adding openshift support. 

We are taking a bit of a different approach. See #111. Instead of running etcd as privileged container, we improve the pv-hostpath chart to set up all the required permissions and SELinux labels.

To install on openshift, you will then need to run 
```
helm install piraeus-etcd-pv --set "nodes={....}" --set selinux=true
```

And use the following settings for the piraeus chart itself:
```yaml
global:
  setSecurityContext: false
csi-snapshotter:
  # already configured by OCP
  enabled: false
etcd:
  podsecuritycontext:
    supplementalGroups: [1000]
```

Since I'd rather not have etcd run as privileged container, I would prefer our method. Are there any benefits of your method that I missed?

dkhachyan:
Hey! Your method is much better! Thank you! I will try it.

PR #115, opened 2020-11-10
propagate CR updates to created resources

WanzenBug:
Changes to Linstor* CRs will now trigger updates on all created resources.
Any changes made to the resources that are _not_ explicitly set by the
operator will not be touched. Also force a rerun every minute to update the
status field.

The implementation is heavily inspired by how "kubectl apply" works.
The idea is:
* Store the "applied configuration" in an annotation
* Compare wanted with current and applied configuration in a 3-way-merge
* Patch if the 3-way-merge patch is not empty

There are some manual fixes necessary to ensure that the serialized spec
only contains actually set fields:
* metadata.creationTimestamp is serialized as "null", it should be skipped
* status values should be skipped too


JoelColledge (approved):
Looks good to me.

`CreateOrUpdate` is magic :smile: 

If you could in future separate renaming, refactoring and functional changes into separate commits it would make reviewing easier.

PR #130, opened 2020-12-14
Explicit namespace for operator and etcd

dkhachyan:
Hi! Adding namespace to operator templates may be useful, when generating manifests with helm template command.

### DIFF
--- a/charts/piraeus/charts/etcd/templates/service.yaml
+++ b/charts/piraeus/charts/etcd/templates/service.yaml
@@ -1,10 +1,10 @@
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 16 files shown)


WanzenBug (changes requested):
Thank you for this PR!

There are two things missing before merging:
* take a look at `./charts/piraeus/charts/etcd/templates`, it should also use explicit namespaces everywhere. Also the `service.yaml` there has 2 metadata keys for some reason. they can be merged.
* after making your changes, please run `make deploy/piraeus`. This will re-generate the rendered deployment example and the CI checks will pass

WanzenBug (approved):

WanzenBug:
Thanks!

Issue #123, opened 2020-11-20
Error writing controller logs to /var/log/linstor-controller, when running on OpenShift


dkhachyan:
Can I fix this issue without running controller pod with root privileges?

log:
```
time="2020-11-18T19:13:48Z" level=info msg="running k8s-await-election" version=refs/tags/v0.2.0
I1118 19:13:48.722244       1 leaderelection.go:242] attempting to acquire leader lease  piraeus-operator/piraeus-cs...
time="2020-11-18T19:13:51Z" level=info msg="long live our new leader: 'piraeus-cs-controller-5f998d764b-btsjm'!"
I1120 12:17:59.154742       1 leaderelection.go:252] successfully acquired lease piraeus-operator/piraeus-cs
time="2020-11-20T12:17:59Z" level=info msg="long live our new leader: 'piraeus-cs-controller-5f998d764b-zz46s'!"
time="2020-11-20T12:17:59Z" level=info msg="starting command '/usr/bin/piraeus-entry.sh' with arguments: '[startController]'"
LINSTOR, Module Controller
Version:            1.9.0 (678acd24a8b9b73a735407cd79ca33a5e95eb2e2)
Build time:         2020-09-23T09:33:23+00:00
Java Version:       11
Java VM:            Debian, Version 11.0.8+10-post-Debian-1deb10u1
Operating system:   Linux, Version 5.6.19-300.fc32.x86_64
Environment:        amd64, 1 processors, 29694 MiB memory reserved for allocations


System components initialization in progress

12:17:59,493 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
12:17:59,494 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback.groovy]
12:17:59,494 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [file:/usr/share/linstor-server/lib/conf/logback.xml]
12:17:59,549 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - debug attribute not set
12:17:59,551 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [ch.qos.logback.core.ConsoleAppender]
12:17:59,553 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [STDOUT]
[...]

PR #132, opened 2020-12-15
anyuid for controller sa

dkhachyan:
Fix https://github.com/piraeusdatastore/piraeus-operator/issues/123


WanzenBug:
Thank you for this PR!

However, I'd rather not merge this in favour of https://github.com/piraeusdatastore/piraeus/pull/55
With the change to the image the current [openshift config](https://github.com/piraeusdatastore/piraeus-operator/blob/master/charts/piraeus/values-openshift.yaml) works without any additional privileges

dkhachyan:
Thank you!

PR #133, opened 2020-12-16
chart: add HA Controller deployment

WanzenBug:


charts/piraeus/values.yaml
```diff
@@ -82,3 +82,10 @@ operator:
     kernelModuleInjectionImage: quay.io/piraeusdatastore/drbd9-bionic:v9.0.25
     kernelModuleInjectionMode: Compile
     kernelModuleInjectionResources: {}
+HAController:
```
JoelColledge, line 85: Helm values should begin with a lowercase letter https://helm.sh/docs/chart_best_practices/values/

JoelColledge (commented):
I have to get used to reading HA Controller as "High Availability Controller" rather than "Highly Available (Linstor) Controller" :smile:

JoelColledge (approved):

PR #136, opened 2020-12-17
Use a timeout of 1m for all CSI components

WanzenBug:
* Provisioner timeout increased from 15s to 1m
* Other timeouts reduced from 4m to 1m

The provisioner probably also had a timeout of 4 minutes in an older version, but that was changed at some point. 1 Minute seemed to completely remove the "Context cancelled" messages in a test run, while not seeming as excessive as the 4m for the other components


WanzenBug:
@JoelColledge ready. Had to remove the parameter on the `node-driver-registrar` container, seems that was too recent for our image version (and also pretty useless, as we just return static information to that plugin)

JoelColledge (approved):

PR #143, opened 2020-12-29
use strategy type: Recreate for linstor-controller

kvaps:
Hi, `strategy: {"type": "Recreate"}` instructs Kubernetes to delete all existing pods before applying new ones.
I believe that this setting will provide more seamless upgrades for linstor-controller pods.

### DIFF
--- a/pkg/controller/linstorcontroller/linstorcontroller_controller.go
+++ b/pkg/controller/linstorcontroller/linstorcontroller_controller.go
@@ -745,6 +745,7 @@ func newDeploymentForResource(controllerResource *piraeusv1.LinstorController) *
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{MatchLabels: labels},
 			Replicas: controllerResource.Spec.Replicas,
+			Strategy: appsv1.DeploymentStrategy{Type: appsv1.RecreateDeploymentStrategyType},
 			Template: corev1.PodTemplateSpec{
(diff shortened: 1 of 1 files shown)


WanzenBug (commented):
Is there a specific problem you encountered during an upgrade?

From my experience, upgrades work just fine with the default `RollingUpdate` strategy. I can see that there might be issues in the event that:
1. An updated pod starts
2. The old leader gets removed
3. The updated pods gains leader status and updates/migrate the database
4. The updated pod crashes
5. An existing old pod takes over as leader, but fails to start because of the newer database revision.

The end result would be the same, as eventually an updated pod would be leader again, and the outdated pods would be removed.

kvaps:
OK, in this way I don't see any problems with the upgrades except the time spent for this upgrade.

As far as can I know Kubernetes kills the oldest pods before the others during the rolling-upgrade.
So I just afraid the situations when leader will be reelected many times during the upgrade.

WanzenBug:
I just gave `Recreate` a spin, and now I agree that it is an improvement. 

Somehow I was thinking of the way a `StatefulSet` works, with terminating Pods one-by-one. Deployment+Recreate work much faster.

I'm happy to merge this. @kvaps please rebase, then I'll get it merged

kvaps:
sure, rebased

WanzenBug (approved):

WanzenBug:
Thanks!

PR #146, opened 2021-01-07
configurable linux kernel headers location

dkhachyan:
In OpenShift environment(FCOS) kernel headers are not accessible in /usr/src directory
Additionaly /usr/src directory is ReadOnly in FCOS.
We have to download kernel headers inside kernel injector container, during entrypoint, or when building
 docker image(in offline configuration), like
https://gitlab.com/nvidia/container-images/driver/-/blob/master/ubuntu16.04/nvidia-driver

### DIFF
--- a/charts/piraeus/crds/piraeus.linbit.com_linstorsatellitesets_crd.yaml
+++ b/charts/piraeus/crds/piraeus.linbit.com_linstorsatellitesets_crd.yaml
@@ -648,6 +648,13 @@ spec:
               imagePullPolicy:
                 description: Pull policy applied to all pods started from this controller
                 type: string
+              kernelModuleInjectionHeadersLocation:
+                description: KernelModuleInjectionHeadersLocation defines linux kernel
+                  headers location
+                enum:
+                - Container
+                - Host
+                type: string
               kernelModuleInjectionImage:
                 description: kernelModuleInjectionImage is the image (location + tag)
                   for the LINSTOR/DRBD kernel module injector
--- a/charts/piraeus/values.yaml
+++ b/charts/piraeus/values.yaml
@@ -82,3 +82,4 @@ operator:
     kernelModuleInjectionImage: quay.io/piraeusdatastore/drbd9-bionic:v9.0.25
     kernelModuleInjectionMode: Compile
     kernelModuleInjectionResources: {}
+    kernelModuleInjectionHeadersLocation: "Host"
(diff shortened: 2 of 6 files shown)


WanzenBug:
Hey! Any particular reason this was closed? At first glance it looks useful.

dkhachyan:
Ok, then i'll try to fix validation errors... have you any recommendations
after review?

WanzenBug (changes requested):
Re validation-errors: We use [`pre-commit`](https://pre-commit.com/) to validate each PR. If you install that it will run the necessary validation before you commit.

So I just had to think about the exact use again: We are running on a host without kernel headers, which means they have to be packaged with the image already. But that means the image could already contain the logic to set the header location, right?

dkhachyan:
When we are running on a host without kernel headers we can use a scenario like this:
- Disable mounting /usr/src
- Query kernel version in entry.sh
- Download kernel headers in entry.sh
- when offline, we can download kernel headers during image build

When /usr/src mount is enabled, it overwrites everything in /usr/src when pod starts
As far as I no, almost all linux distributions installs headers exactly in /usr/src, and we can not change this directory during package install. May be we can move kernel headers into any other dir after package installation and use KDIR environment variable, but this can only be done during image build.

WanzenBug:
The problem I see is with:

> * Download kernel headers in entry.sh

This is heavily dependent on which distribution you are using. We could make it work for just FCOS, but I'd rather not special-case another distribution.

If we want to make it work for offline environments, it would not work anyways. So how about we instead
Provide instructions on how to build your own image. With that I mean something like:
```
FROM fedora:33 as kernel-src
RUN dnf update && dnf install -y kernel-devel

FROM quay.io/piraeusdatastore/drbd9-centos8:v9.0.27-1
COPY --from=kernel-src /usr/src /opt/kernel-src
ENV KDIR=/opt/kernel-src
```
This image can then replace the default image, and it should use the packaged kernel sources rather than reading from `/usr/src`
I think this would be "good enough"

dkhachyan:
OK, thanks!

WanzenBug:
@dkhachyan did you manage to install DRBD on FCOS? During #149 I tried adding a section which describes the above Dockerfile method. During that I noticed:
* There is no matching `kernel-devel` package available for the kernel used by FCOS
* The one `kernel-devel` package that was available does not seem to be compatible with DRBD from my short tests

So I was wondering if you managed to get something running?

dkhachyan:
@WanzenBug, sorry for the long answer. I use koji, to install kernel headers in FCOS

```
RUN dnf install -y \
        koji  && \
        koji download-build --rpm --arch=x86_64 kernel-devel-${KERNEL_VERSION} && \
        dnf install -y kernel-devel-${KERNEL_VERSION}.rpm && \
        rm -f kernel-devel-${KERNEL_VERSION}.rpm && \
    rm -rf /var/cache/yum/*
```

PR #157, opened 2021-01-22
truncate hostname in Job name

tnaganawa:
When hostname is long, Job name also could be longer than 63 characters and Job creation fail.
Add trunc to avoid this.

Signed-off-by: Tatsuya Naganawa <tatsuyan201101@gmail.com>


WanzenBug (approved):
This will still fail if the release name is longer than 12 characters. Since we implicitly depend on that name not being too long in lots of places, I am fine with just ensuring that the hostname portion is reasonably short

tnaganawa:
Thanks for the review :)

PR #158, opened 2021-01-25
docs: instructions for custom injector image

WanzenBug:
See the discussion in #146

### DIFF
--- a/doc/host-setup.md
+++ b/doc/host-setup.md
@@ -28,6 +28,59 @@ You can set the image by passing `--set operator.satelliteSet.kernelModuleInject
 [InitContainer]: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
 [#137]: https://github.com/piraeusdatastore/piraeus-operator/issues/137
 
+### Injector image for compiling without headers on host
+
+Installing the kernel headers is not always feasible on the host. A prime example is Fedora CoreOS (FCOS), which uses
+older versions of the mainline Fedora kernel. Often, the matching kernel-headers are no longer available to install and
+manually manipulating the `ostree` is cumbersome. In this section we will show you how to build your own injector image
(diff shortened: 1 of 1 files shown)


doc/host-setup.md
```diff
@@ -28,6 +28,49 @@ You can set the image by passing `--set operator.satelliteSet.kernelModuleInject
 [InitContainer]: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
 [#137]: https://github.com/piraeusdatastore/piraeus-operator/issues/137
 
+### Build your own injector image
```
JoelColledge, line 31: From this title and the introductory paragraph I had the impression that this would be an image containing pre-built DRBD packages. These instructions are actually for building an injector image that fetches the kernel headers automatically. What should name it? `Build an injector image for compiling without headers on host` is a bit long. `Fedora CoreOS: Build an injector image` since that is its primary purpose?

doc/host-setup.md
```diff
@@ -28,6 +28,49 @@ You can set the image by passing `--set operator.satelliteSet.kernelModuleInject
...
+
+If there is no injector image available for your distribution, you can instead build your own. The below example creates
+an injector image for Fedora CoreOS (FCOS).
+
+To build our injector image for FCOS extend the existing CentOS 8 injector image to automatically download and install
+the kernel headers. FCOS does not provide those headers and since FCOS lags behind the normal fedora releases,
+the matching package is not always available from the regular repos.
+
+This script will be the entrypoint to the image. It will download the kernel headers and invoke the DRBD build process:
```
JoelColledge, line 40: Mention that this should be called `entry-override.sh` here to make the steps easier to follow from top to bottom.

JoelColledge (approved):
:+1:

PR #167, opened 2021-03-11
Allow customizing controller env and properties

### LINKED ISSUE #164: Allow setting LINSTOR properties via secrets
It should be possible to:
* Set (controller) properties from our CRDs
* Set "secret" properties, i.e. the value should reside in a kubernetes secret. The property will not be set directly, but via an env variables (needs support from LINSTOR, should be included in the next release)

Proposal:
* Add a `additionalEnv` section to the `LinstorController` CRD. This is concatenated with the environment we already set and
  added to the controller pod template. The value would be an array of env items, just like the `env:` section of a containerspec.
* Add a `additionalProperties` section to the `LinstorController` CRD. It would look like this:
  ```yaml
  additionalProperties:
  - key: StorDriver/Foobar/UsernameEnv
    value: MY_ENV
  ```
  
In conjunction, these features would allow setting a property value from a secret:
```yaml
additionalEnv:
- name: MY_PASSWORD 
  valueFrom:
    secretKeyRef:
      name: my-k8s-secret
      key: password
additionalProperties:
- key: StorDriver/Foobar/PasswordEnv
  value: MY_PASSWORD
```
Note: whats missing in the LINSTOR Controller is the special "*Env" properties. They should be available in the next feature release for selected variables.

WanzenBug:
Adds two new fields to the LinstorController CRD:
spec.additionalEnv, which is appended to the linstor controller env
spec.additionalProperties, which will set the given properties on the
  controller.

Closes #164

### DIFF
--- a/doc/helm-values.adoc
+++ b/doc/helm-values.adoc
@@ -356,6 +356,18 @@ Valid values:: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and
 Description:: Tolerations to pass to the controller pods.
 
 
+=== `operator.controller.additonalEnv`
+Default:: `[]`
+Valid values:: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/[EnvVar list]
+Description:: A list of additional environment variables to pass to the Linstor controller container.
+
+
+=== `operator.controller.additionalProperties`
+Default:: `{}`
+Valid values:: A map with string keys and values
+Description:: A map of properties to set on the Linstor Controller, equivalent to
+calling `linstor controller set-property <key> <value>`
+
 == Piraeus Satellites
 
 === `operator.satelliteSet.enabled`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Additional environment variables and Linstor properties can now be set in the `LinstorController` CRD.
 * Set node name variable for Controller Pods, enabling [k8s-await-election] to correctly set up the endpoint for
   hairpin mode.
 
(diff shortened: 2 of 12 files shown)


doc/helm-values.adoc
```diff
@@ -356,6 +356,18 @@ Valid values:: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and
...
 
+=== `operator.controller.additonalEnv`
+Default:: `[]`
+Valid values:: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/[EnvVar list]
+Description:: A list of additional environment variables to pass to the Linstor controller container.
+
+
+=== `operator.controller.additionalProperties`
+Default:: `[]`
+Valid values:: list of `{ key: k, value: v }` pairs
```
JoelColledge, line 367: Why such pairs and not a yaml map? Is there precedent somewhere?

WanzenBug: I was following kubernetes API conventions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

Reading it again, there is an explicit exception for "pure" maps, so I think we could use a map here, too. I don't really care, I've just gotten used to the kubernetes "everything is a list" approach.

JoelColledge: I would choose a map in this case, since, as you say, it is a "pure" map. Might we ever add additional keys to the objects? Actually, maybe we might want to treat aux properties differently? In that case, we should add that distinction now.

It looks like `/` can appear unquoted in a yaml key, so all the keys should be easily usable in a yaml map.

JoelColledge:
For some reason the check is failing:
```
F0312 07:56:10.838055   10194 deepcopy.go:885] Hit an unsupported type invalid type for invalid type, from ./pkg/apis/piraeus/v1.LinstorCSIDriver
```
:thinking:

WanzenBug:
Yeah, that check has me scratching my head, can't reproduce it locally, can't reproduce it in a fresh docker container... I'm thinking somethings up with the actions cache, but it looks like github still does not allow manually clearing the cache :/

JoelColledge (approved):

PR #170, opened 2021-03-25
reconcile network interfaces in Linstor clusters

WanzenBug:
In some cases, such as a k8s node restart, a pod can change IP address, which
requires updating the net-interface of the node in Linstor.


JoelColledge (approved):
Looks good :+1:

PR #174, opened 2021-04-28
Remove csi node pods from host-network

WanzenBug:
CSI Pods work just fine in the normal pod network. This way they also
play nicer with NetworkPolicies.


JoelColledge (approved):

JoelColledge:
Good point.

PR #175, opened 2021-04-28
chart: update crds for snapshot controller

WanzenBug:
Update the CRDs packaged for the cluster-wide CSI snapshot controller
to v1. v1beta1 CRDs are deprecated and scheduled for removal.


JoelColledge (approved):
I trust that you've just copied these in so they don't need a detailed review.

PR #178, opened 2021-05-06
Add monitoring services

WanzenBug:
Adds monitoring based on drbd-reactor to the satellite set.

If the prometheus-operator is deployed in the cluster, also adds
servicemonitor resources for all satellites and the linstor controller.


pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -46,11 +46,14 @@ import (
 	"github.com/piraeusdatastore/piraeus-operator/pkg/apis/piraeus/shared"
 	piraeusv1 "github.com/piraeusdatastore/piraeus-operator/pkg/apis/piraeus/v1"
 	mdutil "github.com/piraeusdatastore/piraeus-operator/pkg/k8s/metadata/util"
+	"github.com/piraeusdatastore/piraeus-operator/pkg/k8s/monitoring"
 	"github.com/piraeusdatastore/piraeus-operator/pkg/k8s/reconcileutil"
 	kubeSpec "github.com/piraeusdatastore/piraeus-operator/pkg/k8s/spec"
 	lc "github.com/piraeusdatastore/piraeus-operator/pkg/linstor/client"
 )
 
+const DefaultMonitoringImage = "quay.io/piraeusdatastore/drbd-reactor:v0.3.0"
```
JoelColledge, line 55: Our other default images are in a `const.go` file. Shall we do that here too for consistency?

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -199,9 +203,19 @@ func (r *ReconcileLinstorSatelliteSet) reconcileSpec(ctx context.Context, satell
 
 	log.WithField("changed", satelliteCMChanged).Debug("reconcile satellite configmap: done")
 
+	log.Debug("reconcile drbd-reactor configmap")
+
+	drbdReactorCM := mewMonitoringConfigMap(satelliteSet)
```
JoelColledge, line 208: `mew`? :smile:

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -851,6 +887,33 @@ func newSatelliteDaemonSet(satelliteSet *piraeusv1.LinstorSatelliteSet, satellit
 							},
 							Resources: satelliteSet.Spec.Resources,
 						},
+						{
+							Name:            "drbd-prometheus-exporter",
```
JoelColledge, line 891: Do we want to make it optional whether to add this container etc?

WanzenBug: I was thinking about this, too. The problem with that is: how to determine when to turn it on. Since we use Helm, and helm does not replace CRDs, upgrading from an older version would mean you'd not get the benefit of monitoring, unless you take the extra steps to replace the CRDs.

That reminds me, I still need to update the helm values...

JoelColledge: I don't think we should decide based on the behavior of updated systems. There will presumably/hopefully be many more new systems set up in the future than there are existing systems. I would make it optional but enable it in our Helm values.

Enabling it for existing systems would just be a matter of replacing the CRDs and editing the CRs (with Helm), right?

JoelColledge (approved):
Looks good to me.

JoelColledge:
PR is still marked as a draft. Otherwise ready to merge as far as I can see.

PR #190, opened 2021-06-11
Add documentation

boomanaiden154:
I made some additions to the documentation pages based on some of the issues that I have come across. The additions I have made are as follows:
- Added documentation for multipath configuration on Ubuntu 20.04
- Added line in the `storage.md` file in the LVM pool configuration section to indicate that volume snapshots on volumes provisioned through an LVM pool are not supported.

### DIFF
--- a/doc/host-setup.md
+++ b/doc/host-setup.md
@@ -6,6 +6,27 @@ following options to load DRBD:
 
 [DRBD]: https://github.com/LINBIT/drbd/
 
+## Software compatibility
+
+Certain pieces of software conflict with Piraeus and either need to be configured to not interfere with Piraeus or they need
+to be disabled to ensure Piraeus can function properly.
+
+### Multipathd
+
+If you are using `multipathd`, you will most likely need to change some configuration settings to ensure that Piraeus can function
+properly as `multipathd` by default will interfere with Piraeus. `multipathd` by default opens up DRBD devices which prevents Piraeus
+from working properly. To configure `multipathd`, put the following in the file `/etc/multipath/conf.d/drbd.conf`:
+
+```
+blacklist {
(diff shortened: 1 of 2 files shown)


doc/host-setup.md
```diff
@@ -6,6 +6,22 @@ following options to load DRBD:
...
+
+In Ubuntu 20.04, the multipath daemon opens up DRBD devices if it is not configured which prevents Piraeus from working properly.
+To configure the multipath daemon, put the following into the file `/etc/multipath/conf.d/drbd.conf`. You will need to create the
+directory `/etc/multipath/conf.d` on new installations of Ubuntu 20.04.
+
+```
+blacklist {
+        devnode "^drbd[0-9]+"
+}
+```
```
WanzenBug, line 23: I'd rephrase this section a bit: the problem with `multipathd` doesn't only exist on ubuntu 20.04, but basically in any default multipath installation. The difference is that Ubuntu 20.04 has `multipathd` installed and enabled by default.

So in essence I'd change this section to say "If you are using `multipathd`..." and add that Ubuntu 20.04 enables that by default as a note.

WanzenBug (changes requested):
Thank you for improving our docs. I'd suggest to change the section on multipathd a bit.

doc/host-setup.md
boomanaiden154: I just adjusted the section to be labelled "Software compatibility" instead of making it distro specific and I updated the multipath section to account for the changes that you requested. It should be more general now for all multipath installations.

WanzenBug (approved):
Thanks again, changes look good.

Please squash the changes into one commit and use `--signoff` to indicate that the contribution is your own work.

WanzenBug (approved):

PR #203, opened 2021-07-27
Enable csi storage capacity tracking

WanzenBug:
Storage Capacity Tracking can be used to influence Kubernetes default
scheduler when placing Pods with lazy-bound PVCs.

Linstor CSI supports the necessary CSI spec, this just enables the feature
in the csi sidecar.

### DIFF
--- a/charts/piraeus/templates/csi-controller-rbac.yml
+++ b/charts/piraeus/templates/csi-controller-rbac.yml
@@ -93,7 +93,51 @@ roleRef:
   kind: ClusterRole
   name: external-provisioner-runner
   apiGroup: rbac.authorization.k8s.io
-
+---
+kind: Role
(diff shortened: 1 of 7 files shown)


charts/piraeus/templates/csi-controller-rbac.yml
```diff
@@ -93,7 +93,51 @@ roleRef:
   kind: ClusterRole
   name: external-provisioner-runner
   apiGroup: rbac.authorization.k8s.io
-
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-provisioner-capacity-syncer
```
JoelColledge, line 100: Would `csi-capacity-syncer` do? The 2 `*er` words in the name make it a bit of a mouthful.

JoelColledge (approved):

PR #207, opened 2021-08-18
Add github-action and Dockerfile for multi-arch build

toelke:
The need for this change has been discussed in https://github.com/piraeusdatastore/piraeus/issues/74

This change is probably not extremely elegant; I saw that the `operator-sdk`-version used by this repository is quite old. I am circumventing it here by "manually" building the go-code during building the docker image.

I do not expect this to be merged like this, but I want to start a discussion...


WanzenBug:
On the whole, I'm not against removing operator-sdk from the build process. I think that's what happens in newer operator-sdk versions anyways. We are only using such an old versions because nobody has the time to upgrade.

operator-sdk doesn't actually do anything during the build, just calls `go build`. Ideally, we would just run `go build` with `GOARCH=amd64` and once with `GOARCH=arm64`, and copy the artifact into `gcr.io/distroless/static`. I don't think we need anything from debian.

toelke:
I am doing the first part of this using docker build containers -- that seemed easier for me :-|

I will look into putting the result into `distroless` and see what happens.

toelke:
So running distroless (but not static) is working for both arm64 and amd64.

build/Dockerfile.multiarch
```diff
@@ -0,0 +1,10 @@
+FROM golang:buster AS builder
+
+COPY . /build
+WORKDIR /build
+RUN go build -o ./_output/bin/piraeus-operator -gcflags all=-trimpath=. --asmflags all=-trimpath=. github.com/piraeusdatastore/piraeus-operator/cmd/manager
```
WanzenBug, line 5: I suspect the issue with distroless/static is that the go build command doesn't produce a static binary. You might want to use `CGO_ENABLED=0 go build --ldflags '-extldflags "-static"' ...` and try again.

WanzenBug (changes requested):
Overall looks good. I'd still like to try creating a `distroless/static` image and also removing:
* the existing dockerfile + helper scripts
* the existing github action that builds+pushes using operator-sdk

build/Dockerfile.multiarch
toelke: Yes, I saw it was a dynamic binary and short googling did not produce a command that would generate a static binary. `CGO_ENABLED` looks promising, though. Now I'm learning go and github actions. I just wanted storage :-D

toelke: Yes, now it's static.

Makefile
```diff
@@ -1,107 +0,0 @@
...
-OPERATORSDKCMD := GOROOT=$(GOROOT) GOPATH=$(GOPATH) $(OPERATORSDK)
-
-help:
-	@echo "Useful targets: 'update', 'upload'"
-
-all: update upload
-
-.PHONY: update
-update:
-	$(OPERATORSDKCMD) build --image-build-args "--no-cache=$(NOCACHE)" $(PROJECT):$(TAG)
```
WanzenBug, line 22: Don't delete the whole makefile. Just replace this line with `docker build -t $(PROJECT):$(TAG) --no-cache=$(NOCACHE) -f build/Dockerfile.multiarch .`

build/Dockerfile.multiarch
```diff
@@ -0,0 +1,10 @@
+FROM golang:buster AS builder
+
+COPY . /build
+WORKDIR /build
+RUN CGO_ENABLED=0 go build -o ./_output/bin/piraeus-operator --ldflags '-extldflags "-static"' -gcflags all=-trimpath=. --asmflags all=-trimpath=. github.com/piraeusdatastore/piraeus-operator/cmd/manager
+
+FROM gcr.io/distroless/static-debian10:latest
```
WanzenBug, line 7: Please switch to `gcr.io/distroless/static:latest` and add a `USER 1001` line to keep the old default uid.

WanzenBug (changes requested):

toelke:
About the failing pre-commit hook that checks for a correkt `operator-sdk`-setup: Should I remove the hook or should I rename the `Dockerfile`?

WanzenBug:
> About the failing pre-commit hook that checks for a correkt `operator-sdk`-setup: Should I remove the hook or should I rename the `Dockerfile`?

Please rename the dockerfile, we still need a working operator-sdk for some things.

toelke:
Thank you so much for your patience and all the good tips!

Makefile
WanzenBug: Needs to changed to `-f build/Dockerfile` now

WanzenBug:
Almost ready to merge, just missing the one fix on the Makefile. Already gave it a test run, seems to be working without issue, and we have a smaller image too :+1:

WanzenBug (approved):

WanzenBug:
:tada:

toelke:
Thanks again

PR #209, opened 2021-08-20
Use "nonroot" user instead of hardcoded uid

WanzenBug:
`gcr.io/distroless/static` has a predefined "nonroot" user with a
dedicated home directory. While not strictly needed, it's probably
still a good idea to use one of the provided user instead of hard-coding
a random uid.

Note: `gcr.io/distroless/static` was added in #207


JoelColledge (approved):
Code quality improvement :+1:

PR #216, opened 2021-09-06
copy labels from k8s node to satellite properties

WanzenBug:
Automatically apply all node labels as auxiliary properties, ready
to be used during volume scheduling (see replicasOnSame and
replicasOnDifferent in the CSI driver).


JoelColledge (approved):
Looks great. Features like this make the operator worthwhile.

Issue #204, opened 2021-08-09
Ping problems for piraeus-operator-etcd (missing setcap NET_RAW ?)


Dr-Shadow:
Hello, I'm having issues recently to deploy piraeus-operator on a self hosted kubernetes 1.22 from scratch.
For tests, I used etcd with root user (I removed securitycontext for the etcd instances).
It seems etcd is missing some capabilites in the security context unless I'm mistaken.

➜  infra git:(master) ✗ kubectl logs -n piraeus-system piraeus-operator-etcd-0
[...]
Waiting for piraeus-operator-etcd-0.piraeus-operator-etcd to come up
ping: socket: Address family not supported by protocol
[...]

➜  infra git:(master) ✗ kubectl exec -it -n piraeus-system piraeus-operator-etcd-0 -- whoami
root

➜  infra git:(master) ✗ kubectl exec -it -n piraeus-system piraeus-operator-etcd-0 -- ping -4 google.fr
ping: socket: Operation not permitted
command terminated with exit code 2

1 master / 3 nodes Kubernetes 1.22 on Ubuntu 21.04 / CRI-O

WanzenBug:
Hello, thanks for reporting this issue.

Have you changed anything else on your system? I was under the impression that `CAP_NET_RAW` is on by default. But maybe that changed in recent ubuntu/cri-o versions?

Dr-Shadow:
CRI-O version is 1.21.2

I will dig into this issue the next days, I got piraeus working probably by adding setcap NET_RAW. I can't confirm this yet because this "could" have been a network issue.

For anyone wondering where I could set etcd capabilities and without rootless for debug I got this :

```
diff --git a/charts/piraeus/charts/etcd/templates/statefulset.yaml b/charts/piraeus/charts/etcd/templates/statefulset.yaml
index c28fafe..eae5275 100755
--- a/charts/piraeus/charts/etcd/templates/statefulset.yaml
+++ b/charts/piraeus/charts/etcd/templates/statefulset.yaml
@@ -50,15 +50,12 @@ spec:
 {{- if .Values.tolerations }}
       tolerations:
 {{ toYaml .Values.tolerations | indent 8 }}
-{{- end }}
-      securityContext:
-{{- if .Values.podsecuritycontext }}
-{{ toYaml .Values.podsecuritycontext | indent 8 }}
-{{- else }}
-{{ include "operator.defaultpodsecuritycontext" $ | indent 8 }}
[...]

PR #217, opened 2021-09-06
Add NET_RAW capability on etcd container securitycontext

Dr-Shadow:
Related to https://github.com/piraeusdatastore/piraeus-operator/issues/204

Not tested under containerd but it should be working since containerd doesn't drop NET_RAW by default

The capability should be added on container that's why I'm not using the configurable value PodSecurityContext (see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podsecuritycontext-v1-core)

### DIFF
--- a/charts/piraeus/charts/etcd/templates/statefulset.yaml
+++ b/charts/piraeus/charts/etcd/templates/statefulset.yaml
@@ -59,6 +59,9 @@ spec:
 {{- end }}
       containers:
       - name: {{ template "etcd.fullname" . }}
+        securityContext:
+          capabilities:
+            add: ["NET_RAW"]
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.global.imagePullPolicy | quote }}
         ports:
--- a/deploy/piraeus/charts/etcd/templates/statefulset.yaml
+++ b/deploy/piraeus/charts/etcd/templates/statefulset.yaml
@@ -47,6 +47,9 @@ spec:
         fsGroup: 1000
       containers:
       - name: piraeus-op-etcd
+        securityContext:
+          capabilities:
+            add: ["NET_RAW"]
         image: "gcr.io/etcd-development/etcd:v3.4.15"
         imagePullPolicy: "IfNotPresent"
         ports:


WanzenBug:
Thanks for getting to the bottom of this!

Would you be willing to improve this a bit further by:

* dropping all capabilities
* adding only those that are actually needed.

That should prevent similar issues in the future.
You might also need to run `make deploy/piraeus` once to sync the example deployment with your changes to the template.

As a side node: seems my test setup using kubespray overrides the cri-o default capabilities with their own, which is why I couldn't reproduce the issue.

Dr-Shadow:
Another way to fix it would be to remove the ping commands there (so it means we have to find another way to check if pod is ready) :

https://github.com/piraeusdatastore/piraeus-operator/blob/master/charts/piraeus/charts/etcd/templates/statefulset.yaml#L183

https://github.com/piraeusdatastore/piraeus-operator/blob/master/charts/piraeus/charts/etcd/templates/statefulset.yaml#L191

So that the default capabilities should be enough.

I'm wondering why we aren't using some official etcd chart like the bitnami one ? https://artifacthub.io/packages/helm/bitnami/etcd

WanzenBug:
> Another way to fix it would be to remove the ping commands there

I already had a look through the available commands in the "official" etcd image, and ping seemed to be the only reasonable. You might be able to use an init-container for the same purpose, but that badly interacts with the readiness state and DNS resolution I think. Would have to be tried

> I'm wondering why we aren't using some official etcd chart like the bitnami one

Reason is given in https://github.com/piraeusdatastore/piraeus-operator/commit/7c35509df81819d270f2788c74ca5dc8f7cbcbee
TLDR: we had issues that a simple node failure wasn't recoverable. The replacement seemed to solve those issue.

But I agree in principle. It would be easier if we didn't have to maintain an Etcd chart. There is some ongoing work in LINSTOR to directly use the K8s API as persistent storage (like calico for example). But that will take time, so until then we need Etcd.

Dr-Shadow:
Ok since I would have to test at least these capabilities, I would just stick with the default ones provided by CRI-O / containerd unless we have to add specific ones.
Especially since I'm not an etcd expert, I can't tell if capabilities are used or not.

From `crio config --default`
```
# List of default capabilities for containers. If it is empty or commented out,
# only the capabilities defined in the containers json file by the user/kube
# will be added.
default_capabilities = [
        "CHOWN",
        "DAC_OVERRIDE",
        "FSETID",
        "FOWNER",
        "SETGID",
        "SETUID",
        "SETPCAP",
        "NET_BIND_SERVICE",
        "KILL",
]
```

WanzenBug (dismissed):
Looks good to me, LGTM. 

Only thing missing is to run `make deploy/piraeus` once and commit those changes. Then CI should pass and we can merge this.

Dr-Shadow:
> Only thing missing is to run `make deploy/piraeus` once and commit those changes. Then CI should pass and we can merge this.

Oh that's why the tests were failing, good to know

WanzenBug (approved):

PR #219, opened 2021-09-08
enable CSI topology and disable STORK by default

WanzenBug:
STORK can often be replaced by using a StorageClass which sets the volume
binding mode to "WaitForFirstConsumer". For optimal placement, CSI needs
to run with Topology support, so we enable that by default.

Related: https://github.com/piraeusdatastore/linstor-csi/pull/132


CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Automatically add labels on Kubernetes Nodes to LINSTOR satellites as Auxiliary Properties. This enables using
   Kubernetes labels for volume scheduling, for example using `replicasOnSame: topology.kubernetes.io/zone`.
 
+### Changed
+
+- Enable CSI topology by default, allowing better volume scheduling with `volumeBindingMode: WaitOnFirstConsumer`.
```
JoelColledge, line 18: In [the docs](https://kubernetes.io/docs/concepts/storage/storage-classes/) I only find `WaitForFirstConsumer`.

JoelColledge (commented):
Have you discussed this with others? It seems very reasonable now that the CSI Topology feature has GA status. But it feels like quite a significant change and I'm not familiar with the pros and cons of CSI Topology vs Stork myself.

CHANGELOG.md
WanzenBug: good catch

WanzenBug:
> Have you discussed this with others?

Not directly up until now. I agree it's a major change.

From my point of view:

#### CSI
##### Pros
* It's included in the spec
* It doesn't require additional components
* It enables optimal volume placement even if a pod is using multiple PVCs. Using `WaitForFirstConsumer` all PVCs will be placed on the same node.
##### Cons
* If a Pod is moved, it doesn't have the same insight as STORK to chose from one of the available replicas.
* There might be some additional load on the API servers (see https://github.com/piraeusdatastore/linstor-csi/issues/54)

#### Stork
##### Pros
* Allows smart placement of pods even for existing volumes
* Can differentiate between diskfull and diskless nodes

##### Cons
* It's another component we need to deploy and keep up to date
* It currently doesn't work on Kubernetes 1.22
* It (to my knowledge) doesn't work if just one PV was created using `WaitForFirstConsumer`
* It can't influence the placement of replicas directly, only try to place pods on the "right" nodes. I.e. you could end up with a node with many pods, while another node sits idle simply because one node has "better" storage.

In any case, this is not urgent to merge, so I'd invite any additional feedback on this topic (/cc @rck , @alexzhc )

rck (approved):
my gut feeling is to go for it, there are more advantages.

JoelColledge (approved):
Looks like the way forward :+1:

PR #235, opened 2021-11-08
Evict and remove nodes with no associated K8s Node

WanzenBug:
In case all of the below is true for a LINSTOR node:

* Satellite is offline
* Satellite was registered by controller
* Kubernetes API does not have a node of the same name

we evict + remove the node from LINSTOR.

Some cloud providers handle upgrades by spinning up new nodes, then purging
the old ones. Using evict ensures that after an upgrade, we still have the
expected number of replicas on the new nodes.

### DIFF
--- a/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
+++ b/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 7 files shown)


WanzenBug:
Note: A better implementation would add a finalizer on the kubernetes node, run the currently non-existant "node evacuate" command and remove the finalizer once complete. I guess that is something for Operator v2. The list keeps getting longer :/

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -398,53 +399,65 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAllNodesOnController(ctx context
...
-			defer close(output)
-			output <- r.reconcilePod(ctx, linstorClient, satelliteSet, pod)
-		}()
+		g.Go(func() error {
+			return r.reconcilePod(gCtx, linstorClient, satelliteSet, pod, k8sNode)
+		})
 	}
 
-	logger.Debug("start collecting per-node reconciliation results")
+	err = g.Wait()
```
JoelColledge, line 445: This only returns the first error, hiding the others. Previously we collected them all. Do we want this?

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -1331,6 +1337,67 @@ func (r *ReconcileLinstorSatelliteSet) controllerReachable(ctx context.Context,
 	return err
 }
 
+// removeDanglingSatellites removes satellites that were registered by the operator and are no longer present.
+func (r *ReconcileLinstorSatelliteSet) removeDanglingSatellites(ctx context.Context, linstorClient *lc.HighLevelClient, k8sNodes []corev1.Node) error {
+	lnodes, err := linstorClient.Nodes.GetAll(ctx, &lapi.ListOpts{
+		Prop: []string{fmt.Sprintf("%s=%s", kubeSpec.LinstorRegistrationProperty, kubeSpec.Name)},
+	},
+	)
```
JoelColledge, line 1345: This looks odd. Close them both on one line since they were both opened on one line?

JoelColledge:
This version collects all errors but doesn't preserve the order (if I'm reading the code right), something that you yourself implemented :thinking:

WanzenBug:
> This version collects all errors but doesn't preserve the order (if I'm reading the code right), something that you yourself implemented thinking

:sweat_smile: Thanks for reminding me of that. I'll just do a boring sort by string value....

JoelColledge (approved):

PR #238, opened 2021-11-15
Release 1.7.0-rc.1

WanzenBug:
Prepare 1.7.0-rc.1. Will remain a release candidate until some known issues in LINSTOR are resolved.

### DIFF
--- a/charts/piraeus/Chart.yaml
+++ b/charts/piraeus/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: piraeus
 description: A Helm chart for Piraeus Operator
 type: application
-version: 1.6.0
-appVersion: 1.6.0
+version: 1.7.0-rc.1
+appVersion: 1.7.0-rc.1
 home: https://piraeus.io/
 dependencies:
   - name: "etcd"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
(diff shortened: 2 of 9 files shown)


charts/piraeus/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: piraeus
 description: A Helm chart for Piraeus Operator
 type: application
-version: 1.6.0
-appVersion: 1.6.0
+version: 1.7.0-rc.1
+appVersion: 1.16.0
```
JoelColledge: We usually only change these in the release commit.

WanzenBug: I plan to to a 1.7.0-rc.1tag, so that people can start testing, even if it's not 100% ready yet.

JoelColledge: I was actually just commenting on the fact that these are also touched in the "update container images" commit. GitHub doesn't show that very clearly. Wholly agree that we should update the versions for `1.7.0-rc.1`.

WanzenBug: Thanks for pointing that out. Removed that part of the automation. appVersion tracking LINSTOR version seems the most sensible to me, since the operator is already basically tied to the chart.

JoelColledge: I would stick with appVersion defining the operator version. It seems more logical to me since that is the "app" that the chart deploys. Also, we could do chart releases without operator releases, in which case we would only change version. In fact we even did this back in 0.2.1 :joy:

WanzenBug: Done. Now I still need to investigate why our tests say the container won't stop...

WanzenBug: Turns out it actually waited for the container to start again, and 30s was not enough for the test setup. increasing the timeout solved the issue. so nothing blocking this from my side.

JoelColledge (approved):

PR #241, opened 2021-11-17
satellites: allow restarting all pods at once

WanzenBug:
in case of version upgrades, it makes no sense to have a rolling upgrade
were only one node is updated at a time: LINSTOR can't deal with version skew.

we set the "MaxUnavailable" pods to 100%, i.e. all pods can be updated at once.


JoelColledge (approved):

JoelColledge:
Good point!

PR #243, opened 2021-12-01
csidriver: restart node pods if topology labels changed

WanzenBug:
The satellite operator applies k8s node labels as aux properties. These are
then reported by the csi nodes as supported topology keys. This report
only happens once on plugin startup. Later changes are not reflected, leaving
the volume provisioning steps with old or outdated topology keys.

The only way to force a re-sync of those keys is to delete pod. This change
compares the LINSTOR Node auxilliary properties with the keys reported by
the csi plugin registration. If keys are present in LINSTOR but not
in the registration, the registration and pod is deleted. After the daemonset
restarts the pod, kubelet will trigger a new registration, making the updated
keys available.

### DIFF
--- a/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
+++ b/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
@@ -22,8 +22,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
+	lapiconst "github.com/LINBIT/golinstor"
+	lapi "github.com/LINBIT/golinstor/client"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -340,18 +345,181 @@ func (r *ReconcileLinstorCSIDriver) reconcileStatus(ctx context.Context, csiReso
...
+		if !mdutil.SliceContains(csiDriver.TopologyKeys, expectedKey) {
+			logger.WithField("topologyKey", expectedKey).Debug("key missing in exported topology keys")
+
+			hasAllKeys = false
+
+			break
+		}
+	}
+
+	if hasAllKeys {
```
JoelColledge, line 474: This looks like it only checks that all linstor node properties with the matching keys are also present in `csiDriver.TopologyKeys`. So it will trigger an update if a new linstor node property is added. What happens if a property is removed or the value changed?

WanzenBug: It's complicated: the `csiDriver.TopologyKeys` is only a list of keys. This list is created by the kubelet when requesting the topology labels on startup. the actual values are directly passed to the k8s node. the `csiDriver.TopologyKeys` is then used by the csi-provisioner to fetch the actual labels from the k8s node.

So for removed labels: they are not present on the k8s node, so we should not have to do anything. For changed values: same idea, the k8s nodes labels are always the source of truth for the value.

JoelColledge: Ok. Yes, that's complicated. I'm fine with how it is. No need to make allowances for possible future behavior changes.

JoelColledge (approved):

PR #252, opened 2022-01-03
monitoring: configure TLS for ServiceMonitor

### LINKED ISSUE #251: Securing linstor API broke Prometheus ServiceMonitor of linstor Controller
There is a problem in ServiceMonitor created by LinstorController if API secured.
A LinstorController creates ServiceMonitor for a Controller by [code](https://github.com/piraeusdatastore/piraeus-operator/blob/1b1e30d2e74eecb8ea7af563b25ee18ad63495f0/pkg/controller/linstorcontroller/linstorcontroller_controller.go#L220):
```
if !controllerResource.Spec.SslConfig.IsPlain() {
    serviceMonitor.Spec.Endpoints[0].Scheme = "https"
}
```
Using only https scheme wrong by:
1. A Controller HTTPS server does ssl client certificate check. So Prometheus can't  connect to Controller without valid ssl client cert.
```
$ curl -k https://piraeus-op-cs.piraeus.svc:3371/
curl: (52) Empty reply from server
```
2. A Prometheus connects to ServiceMonitor endpoint by IP, not by name:
```
"lastError": "Get \"https://10.130.0.53:3371/metrics\": x509: cannot validate certificate for 10.130.0.53 because it doesn't contain any IP SANs",
```
Right way is using ServiceMonitor.spec.endpoints[].tlsConfig.
For me patch is:
```
$ kubectl apply -f - <<EOF
kind: ServiceMonitor
apiVersion: monitoring.coreos.com/v1
metadata:
  name: piraeus-op-cs
  namespace: piraeus
spec:
  endpoints:
  - bearerTokenSecret:
      key: ""
    interval: 30s
    path: /metrics
    port: piraeus-op-cs
    scheme: https
    tlsConfig:
      serverName: piraeus-op-cs.piraeus.svc
      keySecret:
        key: client.key
        name: http-client
      cert:
        secret:
          key: client.cert
          name: http-client
      ca:
        secret:
          key: ca.pem
          name: http-client
EOF
```
Environment:
- OpenShift version 4.9.0-0.okd-2021-12-12-025847 (Kubernetes Version: v1.22.1-1824+934e08bc2ce38f-dirty)
- Helm chart piraeus-1.7.0 (piraeus-op --namespace piraeus --set linstorHttpsClientSecret=http-client,linstorHttpsControllerSecret=http-controller)

WanzenBug:
Prometheus requires special configuration when HTTPS is enabled on the
Controller.

This also fixes an issue caused by missing patch metadata for the
ServiceMonitor resource: Instead of patch, the resource is re-created if
this error is encountered

Fixes #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Use correct secret name when setting up TLS for satellites
+- Correctly configure ServiceMonitor resource if TLS is enabled for LINSTOR Controller.
 
 ## [v1.7.0] - 2021-12-14
 
--- a/pkg/controller/linstorcontroller/linstorcontroller_controller.go
+++ b/pkg/controller/linstorcontroller/linstorcontroller_controller.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	lapi "github.com/LINBIT/golinstor/client"
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	awaitelection "github.com/linbit/k8s-await-election/pkg/consts"
 	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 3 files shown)


WanzenBug:
@OleksandrShtepa would it be possible for you to test this change before we merge it?

OleksandrShtepa:
> @OleksandrShtepa would it be possible for you to test this change before we merge it?

Ok, will try

OleksandrShtepa:
Builded image from **https://github.com/WanzenBug/piraeus-operator.git** branch **https-service-monitor**.
```
helm upgrade --set operator.image=image-registry.openshift-image-registry.svc:5000/piraeus/piraeus-operator:https-service-monitor
```
Got same error:
> failed to reconcile servicemonitor definition: failed to apply patch: the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml

I have deleted ServiceMonitor/piraeus-op-cs, after some time the ServiceMonitor has been created with tlsConfig and Prometheus successfully started to scrape metrics.
I tried to edit some in the ServiceMonitor resource, each time had same error.
So:
- tlsConfig in ServiceMonitor works well
- OnPatchErrorRecreate - does not work

WanzenBug:
Thank you for testing! 

I probably also needs to update the conditions on which the recreate logic triggers. I think right now it only checks for a specific return code created when the API server denies updating the resource... But now we send a "wrong" patch because the ServiceMonitor resource for some reason can't be patched with a strategic merge patch :roll_eyes:

JoelColledge (approved):
Looks good.

PR #253, opened 2022-01-10
CSI: add secrets RBAC for sidecar containers

WanzenBug:
To support secret references in storage and snapshot classes, we need
to allow the external-snapshotter and external-provisioner sidecars access
to secrets.

This commit also adds the "--extra-create-metadata" flag to the external
provisioner, which adds parameters containing the PVC name + namespace
to the CreateVolume call, meaning LINSTOR CSI can optionally create
volume names based on PVC name+namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow the external-provisioner and external-snapshotter access to secrets. This is required to support StorageClass
+  and SnapshotClass [secrets](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html).
+- Instruct external-provisioner to pass PVC name+namespace to the CSI driver, enabling optional support for PVC based
+  names for LINSTOR volumes.
+
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


JoelColledge (commented):
It looks like this grants access to all secrets in all namespaces. Is that right? According to the docs you linked, it does look like the intention is that arbitrary secrets can be used. What if a user only wants to use secrets from particular namespaces and not expose others? Is this a situation we care about?

WanzenBug:
Yes, it grants access to all secrets. And yes, I'm not really happy about that, either. I think the idea is that the StorageClass/SnapshotClass is something that only the administrator can configure, they are cluster wide resources after all.

So there is not really a point to restricting the secrets further. That of course assumes that the admin doesn't configure some insecure (templated) value.... :shrug:

JoelColledge (approved):
There is also the issue of damage limitation if one of the provisioners is compromised. But I guess someone with control over a provisioner can already intercept all access to persistent data, so this isn't that much of an escalation.

WanzenBug:
From a security perspective the CSI plugin itself is actually even worse. For it to work it has to have access to `/var/lib/kubelet/pods` which includes accesing _every_ volume of _every_ pod on that node. One volume basically any pod has: the one with it's service account token. So if you have a malicious CSI plugin, you get access to the whole cluster for basically free.

Best not to think about that....

PR #254, opened 2022-01-11
allow setting components log level via CRs

WanzenBug:
Due to the way the operator reconciles resources, it was not possible to set
the log level for components while the operator was running. With this commit
the CRDs have a new "logLevel" field, which can be used to set a blanket log
level for the component.

If no value is given, it defaults to INFO. This was already the default for
LINSTOR Controller and Satellite. Linstor CSI was previously on DEBUG level,
which was often too verbose, running into storage issues after long uptimes.


JoelColledge (approved):
Looks good to me.

PR #255, opened 2022-01-13
reduce load on satellites

WanzenBug:
* Use cached storage pool information to avoid triggering a rescan of pools.
* Ensure device scan is only executed if storage pools have to be configured.

### DIFF
--- a/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
+++ b/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
@@ -530,7 +530,16 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAutomaticDeviceSetup(ctx context
 		"Op":        "reconcileAutomaticDeviceSetup",
 	})
 
-	logger.Debug("check for re-used device paths")
+	logger.Debug("get existing storage pools on node")
+
+	cached := true
+
+	existingPools, err := linstorClient.Nodes.GetStoragePools(ctx, pod.Spec.NodeName, &lapi.ListOpts{Cached: &cached})
+	if err != nil {
+		return fmt.Errorf("failed to list existing storage pools: %w", err)
+	}
+
+	logger.Debug("check for re-used device paths or existing storage pools")
 
 	devsToConfigure := sets.NewString()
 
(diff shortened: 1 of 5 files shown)


pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -530,20 +530,43 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAutomaticDeviceSetup(ctx context
...
 	devsToConfigure := sets.NewString()
 
+outer:
 	for _, poolConfig := range satelliteSet.Spec.StoragePools.AllPhysicalStorageCreators() {
 		if devsToConfigure.HasAny(poolConfig.GetDevicePaths()...) {
 			return fmt.Errorf("a device referenced in the storage pools is referenced twice")
 		}
 
+		for i := range existingPools {
+			if existingPools[i].SharedSpace == poolConfig.GetName() {
```
JoelColledge, line 553: Is `SharedSpace` what we want to check here? That is intended as a special feature when multiple pools are actually backed by the same physical storage.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -530,20 +530,43 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAutomaticDeviceSetup(ctx context
...
 
 	devsToConfigure := sets.NewString()
 
+outer:
 	for _, poolConfig := range satelliteSet.Spec.StoragePools.AllPhysicalStorageCreators() {
 		if devsToConfigure.HasAny(poolConfig.GetDevicePaths()...) {
 			return fmt.Errorf("a device referenced in the storage pools is referenced twice")
 		}
 
+		for i := range existingPools {
```
JoelColledge, line 552: Maybe factor this out into a function to avoid the need for the labelled loop.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
WanzenBug: Good catch. Should be `StoragePoolName` obviously.

JoelColledge (approved):

PR #257, opened 2022-01-14
drbd-injection: make source directory configurable

WanzenBug:
Some distributions don't use /usr/src to store kernel sources. Notably,
Flatcar Container Linux does not have that directory (and it can't easily be
created, either).
    
This commit adds a new value to the satellite CRD:
    
  kernelModuleInjectionAdditionalSourceDirectory
    
If set, the injection container will be started with the directory bound
from the host (read-only). If not set, /usr/src will be bound to keep
existing behaviour. If the given value is not an absolute path, no directory
from the host will be bound (except for /lib/modules, which is required in
all cases).
    
As a small improvement, we also don't try to mount this directory if
DRBD should not be compiled.

### DIFF
--- a/charts/piraeus/templates/operator-satelliteset.yaml
+++ b/charts/piraeus/templates/operator-satelliteset.yaml
@@ -20,6 +20,9 @@ spec:
   kernelModuleInjectionMode: {{ .Values.operator.satelliteSet.kernelModuleInjectionMode | quote }}
   kernelModuleInjectionImage: {{ .Values.operator.satelliteSet.kernelModuleInjectionImage | quote }}
   kernelModuleInjectionResources: {{ .Values.operator.satelliteSet.kernelModuleInjectionResources | toJson }}
+  {{- if .Values.operator.satelliteSet.kernelModuleInjectionAdditionalSourceDirectory }}
+  kernelModuleInjectionAdditionalSourceDirectory: {{ .Values.operator.satelliteSet.kernelModuleInjectionAdditionalSourceDirectory | quote }}
+  {{- end }}
   {{- if .Values.operator.satelliteSet.storagePools }}
   storagePools:
 {{ toYaml .Values.operator.satelliteSet.storagePools | indent 4 }}
--- a/charts/piraeus/values.yaml
+++ b/charts/piraeus/values.yaml
@@ -83,6 +83,7 @@ operator:
     monitoringImage: quay.io/piraeusdatastore/drbd-reactor:v0.4.4
     kernelModuleInjectionImage: quay.io/piraeusdatastore/drbd9-bionic:v9.1.4
     kernelModuleInjectionMode: Compile
+    kernelModuleInjectionAdditionalSourceDirectory: ""
     kernelModuleInjectionResources: {}
     additionalEnv: []
 haController:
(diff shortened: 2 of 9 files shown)


charts/piraeus/values.yaml
```diff
@@ -83,6 +83,7 @@ operator:
     monitoringImage: quay.io/piraeusdatastore/drbd-reactor:v0.4.4
     kernelModuleInjectionImage: quay.io/piraeusdatastore/drbd9-bionic:v9.1.4
     kernelModuleInjectionMode: Compile
+    kernelModuleInjectionMountUsrSrc: ""
```
JoelColledge, line 86: According to the docs, only `true` and `false` are allowed. Shouldn't this be `true`? How does this even work? :thinking:

charts/piraeus/templates/operator-satelliteset.yaml
```diff
@@ -20,6 +20,9 @@ spec:
   kernelModuleInjectionMode: {{ .Values.operator.satelliteSet.kernelModuleInjectionMode | quote }}
   kernelModuleInjectionImage: {{ .Values.operator.satelliteSet.kernelModuleInjectionImage | quote }}
   kernelModuleInjectionResources: {{ .Values.operator.satelliteSet.kernelModuleInjectionResources | toJson }}
+  {{- if typeIs "bool" .Values.operator.satelliteSet.kernelModuleInjectionMountUsrSrc }}
```
JoelColledge, line 23: Why the extra check here? We don't have it for other values including `dbUseClientCert`, which is also a boolean.

JoelColledge:
Do we want to allow the `/usr/src` path to be customized? It could be useful, but we have no current need for it. The tricky part is that we want the default to be `/usr/src`, but also allow some way of disabling it. Perhaps default to `/usr/src` when empty, and disable it when it is non-empty, but not an absolute path? Then one could specify `none` to disable it.

charts/piraeus/templates/operator-satelliteset.yaml
WanzenBug: See above. If I want to avoid a mandatory CRD upgrade, I have to leave the field out of the CR unless explicitly set. I also can't apply a simple empty check, as that fails when it was explicitly set to true/false (wrong type). I also can't just compare it to a string (type error again). I also can't check if it's false-y, as that would mean an explicit `false` would not be set :/

WanzenBug:
> Do we want to allow the /usr/src path to be customized?

:bulb: why didn't I think of that. I guess that's a better idea than hardcoding it, with all the weird bool handling....

doc/host-setup.md
```diff
@@ -49,6 +49,15 @@ You can set the image by passing `--set operator.satelliteSet.kernelModuleInject
...
 [#137]: https://github.com/piraeusdatastore/piraeus-operator/issues/137
 
+### Flatcar Container Linux
+
+Flatcar Container Linux requires building DRBD from source. Piraeus by default bind mounts `/usr/src` into the injection
+container, as that is required on most distributions. This will fail on Flatcar Container Linux, as that directory
+does not exist there, and can't easily be created (`/usr` is always read-only).
+
+For this case, you can disable mounting of `/usr/src` by passing
+`--set operator.satelliteSet.kernelModuleInjectionAdditionalSourceDirectory="none"` to Helm.
```
JoelColledge, line 59:
```suggestion
`--set operator.satelliteSet.kernelModuleInjectionAdditionalSourceDirectory=none` to Helm.
```

Should work here.

JoelColledge (commented):
Looks good to me overall. Commit message needs updating.

JoelColledge (approved):

PR #259, opened 2022-01-18
Release 1.7.1

WanzenBug:
* snapshot/provisioner secrets roles
* Add optional metadata to CSI provisioner
* Set log level of LINSTOR components
* Enable customizing the kernel source directory if required
* etcd: add priority class
* fixed TLS setup for satellites
* fixed TLS setup for service monitor


JoelColledge:
Shouldn't this be 1.8 according to semver? New functionality.

WanzenBug:
> Shouldn't this be 1.8 according to semver? New functionality.

Strictly speaking true. just felt like it was only very minor features. If I was going strictly by semver we would be at 6.0.0 or something :grin:

JoelColledge (approved):
We're not strict with our other projects, I guess.

PR #260, opened 2022-01-28
Options to set amount of worker threads for CSI-controller

kvaps:
[external-provisioner](https://github.com/kubernetes-csi/external-provisioner) has 100 worker threads set by default. It seems linstor-server is not capable to handle such a load.

I faced many fancy bugs while testing pvc creating in concurrency:
- https://github.com/LINBIT/linstor-server/issues/268
- https://github.com/LINBIT/linstor-server/issues/269
- https://github.com/LINBIT/linstor-server/issues/270
- https://github.com/LINBIT/linstor-server/issues/271
- https://github.com/LINBIT/linstor-server/issues/272

Setting `--worker-threads=10` solves this problem to me.
But I suppose there are might be users with more or less powerful servers, so I made this parameter configurable

PR #261, opened 2022-01-28
Update CSI images

kvaps:
The CSI images updated to latest versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow setting the number of parallel requests created by the CSI sidecars. This limits the load on the LINSTOR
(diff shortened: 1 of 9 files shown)


WanzenBug (dismissed):
I'll need to do some tests, see if additional rules for RBAC are required, but in general: LGTM.

Please "sign-off" as required by the CNCF/DOC bot.

EDIT: and also run `make deploy/piraeus` to generate example deployment.

WanzenBug (approved):

WanzenBug:
@kvaps I notice some issues related to #260, i.e. the attacher image also wants a `-worker-threads` option instead of `-workers`.

Also, one thing I didn't think about was that the changes in the CRDs break backwards compatibility, i.e. when upgrading you need to manually replace the CRDs first.

I'd push some fixes for both issues on this branch, if you don't mind?

kvaps:
Yeah you're right. Sure, no objections.

WanzenBug:
Finally managed to e2e test it. LGTM

PR #262, opened 2022-01-31
Use `type: kubernetes.io/tls` for certificates

kvaps:
This PR unifies certificates standard used for LINSTOR running in Kubernetes.

No more fancy java tools to generate certificates required.
The certificates now can be generated automatically using Helm or cert-manager.

requires https://github.com/piraeusdatastore/piraeus/pull/97 to get merged first


WanzenBug (changes requested):
In general: very much looking forward to merging this.

First of all, I'll need some time to verify that it's all working as expected.

Some minor changes I'd like to see:
* Increment the etcd subchart version. Doesn't really matter from the functional side, but documents that something was changed.
* Add the changes to the unreleased section of CHANGELOG.md
* Add a warning/guide to UPGRADE.md for people that were already using the old TLS setup.

In a future PR we could even automatically create the certs using helm, but I think this PR is already big enough.

kvaps:
> I'll need some time to verify that it's all working as expected.

I was test everything, except etcd certificates.

> Increment the etcd subchart version

The CRDs description and etcd chart dependeny will be updated. Shouldn't we also bump piraeus chart version?

WanzenBug:
> Shouldn't we also bump piraeus chart version?

I always update that as part of the release process.

WanzenBug:
Just tested it. Works like a charm!

One thing that worries me is that only check that the hostname matches the certificate name happens in the HTTP clients, but I guess that makes it easier to provision certificates :upside_down_face: 

@kvaps can you please add the entries in `CHANGELOG.md` and `UPGRADE.md`, then LGTM.

kvaps:
> @kvaps can you please add the entries in CHANGELOG.md and UPGRADE.md, then LGTM.

Sure, will do.

> One thing that worries me is that only check that the hostname matches the certificate name happens in the HTTP clients

Guess this check is disabled on linstor-server side by design.
I tried to issue certificates without server auth usage, and got an error on linstor-controller trying to connect satellites. Adding/removing any SANs didn't make any sense for LINSTOR.

WanzenBug (dismissed):

WanzenBug:
@kvaps sorry, DCO check failed again...

kvaps:
fixed

WanzenBug (approved):

PR #263, opened 2022-02-01
Autogenerate SSL certificates

kvaps:
This PR adds two new methods to automate SSL certificates generation for both communications: 

```
--set linstorSslMethod=helm --set linstorHttpsMethod=helm
```

```
--set linstorSslMethod=cert-manager --set linstorHttpsMethod=cert-manager
```

by default `manual` mode enabled, so it shouldn't brake oportunity to import your own certificates:

requires https://github.com/piraeusdatastore/piraeus-operator/pull/262 to be merged first

### DIFF
--- a/charts/piraeus/templates/_helpers.tpl
+++ b/charts/piraeus/templates/_helpers.tpl
@@ -45,13 +45,57 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
 
 {{/*
+Controller SSL secret name
+*/}}
+{{- define "controller.sslSecretName" -}}
+  {{- if (eq (.Values.linstorSslMethod | default "manual") "manual") -}}
+    {{- .Values.operator.controller.sslSecret }}
+  {{- else }}
+    {{- .Values.operator.controller.sslSecret | default (printf "%s-control-secret" ( include "operator.fullname" . )) }}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+SatelliteSet SSL secret name
+*/}}
+{{- define "satelliteSet.sslSecretName" -}}
+  {{- if (eq (.Values.linstorSslMethod | default "manual") "manual") -}}
+    {{- .Values.operator.satelliteSet.sslSecret }}
+  {{- else }}
+    {{- .Values.operator.satelliteSet.sslSecret | default (printf "%s-node-secret" ( include "operator.fullname" . )) }}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+Controller HTTPS secret name
+*/}}
+{{- define "controller.httpsSecretName" -}}
+  {{- if (eq (.Values.linstorHttpsMethod | default "manual") "manual") -}}
+    {{- .Values.linstorHttpsControllerSecret }}
+  {{- else }}
+    {{- .Values.linstorHttpsControllerSecret | default (printf "%s-controller-secret" ( include "operator.fullname" . )) }}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+Client HTTPS secret name
+*/}}
+{{- define "client.httpsSecretName" -}}
+  {{- if (eq (.Values.linstorHttpsMethod | default "manual") "manual") -}}
+    {{- .Values.linstorHttpsClientSecret }}
+  {{- else }}
+    {{- .Values.linstorHttpsClientSecret | default (printf "%s-client-secret" ( include "operator.fullname" . )) }}
+  {{- end -}}
+{{- end -}}
+
+{{/*
 Endpoint URL of LINSTOR controller
 */}}
 {{- define "controller.endpoint" -}}
   {{- if .Values.controllerEndpoint -}}
     {{ .Values.controllerEndpoint }}
   {{- else -}}
-    {{- if empty .Values.linstorHttpsClientSecret -}}
+    {{- if empty ( include "client.httpsSecretName" . ) -}}
       http://{{ template "operator.fullname" . }}-cs.{{ .Release.Namespace }}.svc:3370
     {{- else -}}
       https://{{ template "operator.fullname" . }}-cs.{{ .Release.Namespace }}.svc:3371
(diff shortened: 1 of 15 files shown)


charts/piraeus/templates/operator-controller.yaml
```diff
@@ -45,6 +45,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     helm.sh/resource-policy: keep
+    helm.sh/hook: pre-install
+    helm.sh/hook-delete-policy: before-hook-creation
```
kvaps, line 49: up to consider

kvaps:
PR rebased

charts/piraeus/templates/_helpers.tpl
```diff
@@ -44,14 +44,58 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
...
+{{- define "controller.sslSecretName" -}}
+  {{- if (eq (.Values.linstorSslMethod | default "manual") "manual") -}}
+    {{- .Values.operator.controller.sslSecret }}
+  {{- else }}
+    {{- .Values.operator.controller.sslSecret | default (printf "%s-control-secret" ( include "operator.fullname" . )) }}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+StatefulSet SSL secret name
```
WanzenBug, line 59: SatelliteSet instead of StatefulSet

charts/piraeus/templates/operator-controller.yaml
WanzenBug: Is this to prevent overriding the secret on updates?

charts/piraeus/templates/operator-https-secret.yaml
```diff
@@ -0,0 +1,85 @@
...
+metadata:
+  name: {{ template "controller.httpsSecretName" . }}
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/resource-policy": "keep"
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+type: kubernetes.io/tls
+data:
+  {{- with genSignedCert $cn (list "127.0.0.1") (list $cn $altName1 $altName2 "localhost") 3650 $ca }}
```
WanzenBug, line 19: not tested, but should we also add `::1` as IP?

charts/piraeus/templates/operator-issuer.yaml
```diff
@@ -0,0 +1,39 @@
+{{- $fullName := include "operator.fullname" . -}}
+{{- if or (eq .Values.linstorHttpsMethod "cert-manager") (eq .Values.linstorSslMethod "cert-manager") }}
```
WanzenBug, line 2: Is it usual for charts to create their own issuer? I was under the impression that issuers are set up once by the admin?

charts/piraeus/templates/operator-ssl-secret.yaml
```diff
@@ -0,0 +1,82 @@
...
+metadata:
+  name: {{ template "controller.sslSecretName" . }}
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/resource-policy": "keep"
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
+type: kubernetes.io/tls
+data:
+  {{- with genSignedCert $cn (list "127.0.0.1") (list $cn $altName1 $altName2 "localhost") 3650 $ca }}
```
WanzenBug, line 19: Same as above, would adding `::1` work?

charts/piraeus/templates/operator-controller.yaml
kvaps: Yes, it will install secret only during `helm install`, then it will never override it during `helm upgrade` even if it was removed.

charts/piraeus/templates/operator-issuer.yaml
kvaps: I suppose you're talking about kind ClusterIssuers which are mainly used to issue certificates using let's encrypt service.

Since we're using self-signed certificate and own creating own Authority, then we have to create our own issuer, this issuer is namespace scoped and nobody else could use it in the cluster.

charts/piraeus/templates/operator-https-secret.yaml
kvaps: added `::1`, is working fine:

```
X509v3 Subject Alternative Name:
    DNS:piraeus-op-cs, DNS:piraeus-op-cs.linstor, DNS:piraeus-op-cs.linstor.svc, DNS:localhost, IP Address:0:0:0:0:0:0:0:1, IP Address:127.0.0.1
```

```console
$ linstor --controllers 'https://[::1]:3371' c v
linstor controller 1.17.0; GIT-hash: 7e646d83dbbadf1ec066e1bc8b29ae018aff1f66
```

WanzenBug (changes requested):
Please rebase

WanzenBug (changes requested):
For some reason the generated deployment in the CI pipeline does not contain the `LinstorContoller` resource. Could you please try to get the pipeline to pass?

kvaps:
Fixed in 72b76a7d68dd7131466eec0e7365403e0b2cbb40 😃

WanzenBug:
Other then the mysterious disappearance of the `LinstorController` resource, this seems to all work as it should. Looking forward to finally merging it.

WanzenBug:
On a hunch, I removed
```
    helm.sh/hook: pre-install
    helm.sh/hook-delete-policy: before-hook-creation
```
From the generated passphrase secret. Now the `LinstorController` is present again in the deploy directory.

I guess that template should be moved to another file... I'll push a fix shortly.

WanzenBug (approved):

doc/security.md
```diff
@@ -162,3 +162,7 @@ On install, add the following arguments to the helm command:
...
+|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| If you ever delete or change the passphrase secret, the LINSTOR Controller can no longer start with a failure message like `Automatic injection of passphrase failed`. You can force the Controller to start by setting the `luksSecret` value in the `LinstorController` resource to `""`. This will _not_ give you access to encrypted items such as remotes, but it will allow the Controller to start. If you need to recover encrypted values, you need to restore the original secret. |
```
kvaps, line 168: What about the case when original secret was lost? Is there any opportunity to assign the new one?

WanzenBug: Just confirmed by the linstor team: no, you can't assign a new one without knowing the old one.

WanzenBug (approved):

kvaps:
> Other then the mysterious disappearance of the `LinstorController` resource, this seems to all work as it should. Looking forward to finally merging it.

I think the main reason it that valid yaml stream is:

```yaml
---
name: that
---
name: and that
```

but not:

```yaml
name: that
---
name: and that
```

I can take a guess that in this case Helm might consider top yaml as part of the hook of bottom yaml for some reason.

WanzenBug (approved):

WanzenBug:
You are obviously right about the yaml streams. Still, even with that fixed, the "one-file" solution will render the wrong template. Something to keep in mind in the future, but for now I think the workaround is good enough.

doc/security.md
kvaps: So in that case the linstor-cluster must be fully recreated with losing configuration of other (maybe unencrypted devices) 🤔

PR #268, opened 2022-02-03
Fix conflicting names

kvaps:
LinstorSattelliteSet and LinstorController are having conflicting entity names.

Eg. if you install linstorController with name `linstor` and linstorSatelliteSet with name `linstor` operator will be failed to upgrade ownerRefference field for `linstor-config` configmap.

This PR fixes this behavior by adding additional suffix for such resources.

### DIFF
--- a/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
+++ b/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
@@ -1039,7 +1039,7 @@ func getObjectMeta(controllerResource *piraeusv1.LinstorCSIDriver, nameFmt, comp
 		Name:      fmt.Sprintf(nameFmt, controllerResource.Name),
 		Namespace: controllerResource.Namespace,
 		Labels: map[string]string{
-			"app.kubernetes.io/name":       kubeSpec.ControllerRole,
+			"app.kubernetes.io/name":       kubeSpec.CSIDriverRole,
 			"app.kubernetes.io/instance":   controllerResource.Name,
 			"app.kubernetes.io/managed-by": kubeSpec.Name,
 			"app.kubernetes.io/component":  component,
--- a/pkg/controller/linstorcontroller/linstorcontroller_controller.go
+++ b/pkg/controller/linstorcontroller/linstorcontroller_controller.go
@@ -176,6 +176,13 @@ func (r *ReconcileLinstorController) reconcileSpec(ctx context.Context, controll
 		return fmt.Errorf("failed to add finalizer: %w", err)
 	}
 
+	log.Debug("reconcile legacy config map name")
+
+	err = reconcileutil.DeleteIfOwned(ctx, r.client, &corev1.ConfigMap{ObjectMeta: getObjectMeta(controllerResource, "%s-config")}, controllerResource)
+	if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -1011,7 +1011,7 @@ func getObjectMeta(controllerResource *piraeusv1.LinstorCSIDriver, nameFmt, comp
 		Name:      fmt.Sprintf(nameFmt, controllerResource.Name),
 		Namespace: controllerResource.Namespace,
 		Labels: map[string]string{
-			"app.kubernetes.io/name":       kubeSpec.ControllerRole,
+			"app.kubernetes.io/name":       component,
```
WanzenBug, line 1014: Why? We already have the component label for that?

WanzenBug (changes requested):
While I am very much in favour of fixing this issue:

This will leave a lot of resources behind on upgrades. If we move forward with this renaming, we have to make sure all old names are cleaned up

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
kvaps: `csi-node` daemonset has following labels set:
```yaml
app.kubernetes.io/component: csi-node
app.kubernetes.io/instance: linstor
app.kubernetes.io/managed-by: piraeus-operator
app.kubernetes.io/name: csi-controller
```

I found it a bit confusing 🤔

WanzenBug: Oh, sure. should probably be `kubeSpec.CSINodeRole`...

kvaps: Yeah, it is passed as third argument to the function:
https://github.com/piraeusdatastore/piraeus-operator/blob/51a74d33b21eba2b262eea173c0b58dc691dfb06/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go#L305

WanzenBug: :bulb: Ah, makes sense. Thinking some more, shouldn't the name just be "linstor-csi" (or piraeus-csi...) :thinking: 

We already have the component label for the distinction between controller and node.

kvaps:
> we have to make sure all old names are cleaned up
What is the best flow for doing this?

WanzenBug:
> What is the best flow for doing this?

Not really sure. We would need a list of the old resource we expect to exist, and delete them. We should also make sure that those resources have the right owner set, so that we don't try to delete resources that just have a matching name accidentally.

kvaps:
There is a good question do we really need to delete these resources.
They are don't bother us with their existing, and will be automatically removed as any others in case of removing CR

WanzenBug:
>  They are don't bother us with their existing, and will be automatically removed as any others in case of removing CR

That is mostly true for the configmaps, but the servicemonitors could get annoying because they might have an invalid scrape config.

WanzenBug (approved):

WanzenBug:
Added some logic that tries to delete the stale resources. @kvaps anything you want to add?

kvaps (commented):
lgtm

PR #269, opened 2022-02-07
Update stork scheduler for 1.22+

kvaps:
upstream https://github.com/libopenstorage/stork/pull/982

### DIFF
--- a/charts/piraeus/templates/stork-deployment.yaml
+++ b/charts/piraeus/templates/stork-deployment.yaml
@@ -11,22 +11,27 @@ data:
   # of Kubernetes, so please update those as required. The requirement to manually specify this
   # list will go away with Kubernetes v1.10 where it will use the defaults if nothing is
   # specified.
-  policy.cfg: |-
-    {
-      "kind": "Policy",
-      "apiVersion": "v1",
-      "extenders": [
-        {
-          "urlPrefix": "http://{{ .Release.Name }}-stork-service.{{ .Release.Namespace }}.svc.cluster.local:8099",
-          "apiVersion": "v1beta1",
-          "filterVerb": "filter",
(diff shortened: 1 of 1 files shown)


WanzenBug (approved):
I trust that you tested this? TBH I don't really feel like testing stork again :sweat_smile:

charts/piraeus/templates/stork-deployment.yaml
```diff
@@ -11,22 +11,27 @@ data:
...
-          "filterVerb": "filter",
-          "prioritizeVerb": "prioritize",
-          "weight": 5,
-          "enableHttps": false,
-          "nodeCacheCapable": false
-        }
-      ]
-    }
+  stork-config.yaml: |-
+    apiVersion: kubescheduler.config.k8s.io/v1beta3
```
kvaps, line 15: Yeah, small note, `kubescheduler.config.k8s.io/v1beta3` is working with 1.23+, while `kubescheduler.config.k8s.io/v1beta2` is working with 1.22+.

I think we can use `semverCompare` method here, just say me which versions we want to support?

kvaps:
And may be it makes more sense to drop STORK in favor of linstor-scheduler-extender, so it would be nice to see it in official piraeus repository https://github.com/piraeusdatastore/piraeus-operator/issues/270

charts/piraeus/templates/stork-deployment.yaml
WanzenBug: General guide is, we want to support every "stable" release, i.e. those that are listed here: https://www.downloadkubernetes.com/

I'm guessing we can't easily support 1.20 and 1.21?

WanzenBug (approved):

PR #272, opened 2022-02-13
cleanup Chart.lock

skillcoder:
This fix issue with installing piraeus operator via [flux2](https://github.com/fluxcd/flux2)

It looks like this:
```
flux get all -A --status-selector ready=false
# flux-system	helmchart/piraeus-piraeus-op	False	dependency build error: failed to add local dependency 'csi-snapshotter': can't get a valid version for constraint '1.0.0'	        	False
```


WanzenBug (changes requested):
TBH, I forgot this file existed... LGTM. please follow the instructions of the DCO bot so we can merge these changes.

WanzenBug:
Fixed in #276

PR #275, opened 2022-02-15
let components wait for LINSTOR API

WanzenBug:
The HA Controller and CSI Driver can only run if the LINSTOR api is available.
To prevent an initial round of crashed containers, we add init containers with
the only purpose of waiting for the LINSTOR API to be available.

The used linstor-wait-until binary is included in the docker images itself, so
we don't need to add yet another image that needs to be configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- HA Controller and CSI components now wait for the LINSTOR API to be initialized using InitContainers.
+
 ## [v1.8.0-rc.1] - 2022-02-24
 
 ### Added
--- a/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
+++ b/pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
@@ -676,8 +676,21 @@ func newCSINodeDaemonSet(csiResource *piraeusv1.LinstorCSIDriver) *appsv1.Daemon
 		Resources: csiResource.Spec.Resources,
 	}
 
+	linstorWaitNodeInitContainer := corev1.Container{
+		Name:            "linstor-wait-node-online",
+		Image:           csiResource.Spec.LinstorPluginImage,
+		ImagePullPolicy: csiResource.Spec.ImagePullPolicy,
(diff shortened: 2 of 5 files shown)


pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -676,8 +676,21 @@ func newCSINodeDaemonSet(csiResource *piraeusv1.LinstorCSIDriver) *appsv1.Daemon
 		Resources: csiResource.Spec.Resources,
 	}
 
+	linstorWaitNodeInitContainer := corev1.Container{
+		Name:            "linstor-wait-node-online",
+		Image:           csiResource.Spec.LinstorPluginImage,
```
kvaps, line 681: Wouldn't it be better to use separate `linstor-wait-until` image for all components instead of putting its binary into the each component image?

WanzenBug: Another image is another component that needs to be configured, another component that users might need to override/replace. then you have to do this weird song&dance with upgrade CRDs, because that's were they need to be configured, and you might need a default value somewhere in the operator.

so I just skip all that, put it into the containers that need to be downloaded in any case and be done with it.

CHANGELOG.md
```diff
@@ -548,5 +552,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v1.6.0]: https://github.com/piraeusdatastore/piraeus-operator/compare/v1.5.1...v1.6.0
 [v1.7.0]: https://github.com/piraeusdatastore/piraeus-operator/compare/v1.6.0...v1.7.0
 [v1.7.1]: https://github.com/piraeusdatastore/piraeus-operator/compare/v1.7.0...v1.7.1
-[v1.8.0-rc.1]: https://github.com/piraeusdatastore/piraeus-operator/compare/v1.7.1...v1.8.0-rc.1
```
JoelColledge, line 551: This doesn't look deliberate.

JoelColledge (commented):
Looks good overall.

CHANGELOG.md
WanzenBug: Yup, rebase gone wrong :/

JoelColledge (approved):

PR #279, opened 2022-02-28
Use controller-gen instead of operator-sdk

kvaps:
operator-sdk v0.19.4 is quite old and already deprecated, as it generates resources wrong.
I just tried to implement `podTemplate` field and it wasn't working for Kubernetes >1.18

see https://github.com/kubernetes/kubernetes/issues/91395

controller-gen which is used by default in newer operator-sdk versions is working fine.


Makefile
```diff
@@ -32,15 +23,34 @@ upload:
 		docker push $$r/$(PROJECT):latest ; \
 	done
 
+CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+.PHONY: controller-gen
+controller-gen:
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0)
```
WanzenBug, line 29: I believe since go 1.17 you can simply `go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0 ...` from anywhere without adding the dependency to `go.mod`. I'd prefer using that instead of manually implementing "package management" in the makefile

WanzenBug (commented):
Changes generally look good and much less painful than I expected. I'd make one change if it's feasible for everyone to use go 1.17.

WanzenBug:
Also, as a general heads-up, I won't merge this for 1.8.0

Makefile
kvaps: Yeah you're right.
This part was taken from the last operator-sdk generated Makefile.

WanzenBug:
@kvaps I think you need to update the github actions to use go 1.17

WanzenBug (approved):
This looks a lot easier than I would have expected :+1:

Issue #180, opened 2021-05-14
RFC: secure components via kube-rbac-proxy


WanzenBug:
Right now configuring HTTPS on all components is quite cumbersome, or in some cases (monitoring with drbd-reactor) just not possible.

One solution I recently stumbled upon: https://github.com/brancz/kube-rbac-proxy. It's small proxy that integrates with the normal RBAC mechanism of kubernetes to secure endpoints. Using it as a sidecar for the linstor-controller would potentially remove the need for [all these steps](https://github.com/piraeusdatastore/piraeus-operator/blob/master/doc/security.md#configuring-secure-communications-for-the-linstor-api)

To use RBAC we would need to identify which components need a specific kind of access. For example: I believe the csi-node container only `GET`s resources, while the csi-controller needs to `POST` resource groups and definitions.

AntonSmolkov:
Just in case - OpenShift has similar thing out-of-the-box
https://github.com/openshift/oauth-proxy

kvaps:
Hi @WanzenBug, I'm going to implement this, but we have our own rbac-proxy which we run with own configuration, eg:

https://github.com/deckhouse/deckhouse/blob/b8024c00bc0c616cdceee8488bfef523d0f5086a/modules/021-kube-proxy/templates/daemonset.yaml#L106-L138 

Actually we're fine with just adding `sidecars: []` option to the LinstorController and LinstorSatelliteSet resources.
But until I started implementing this, maybe do you have better idea?

WanzenBug:
Very interesting. Having an extra `sidecars: []` option could definitely work. There are a few issues if we want to properly secure the API, as we need to change the current listen address for controller and reactor from `0.0.0.0:...` to `127.0.0.1:...` but only if we have the sidecar rbac proxy.

I want to eventually move to a "Operator v2" that learns from all the mistakes in the current implementation. One of the issues currently is that it is quite difficult to adapt the deployments/daemonsets generated by the operator, so there is a lot of fields in our CRDs that are just "pass-through" (resource, affinity, priorityclass, even service accounts, etc...).

Some ways to address this in a v2:

[...]

PR #280, opened 2022-02-28
Rbac proxy

kvaps:
fixes https://github.com/piraeusdatastore/piraeus-operator/issues/180
requires https://github.com/piraeusdatastore/piraeus-operator/pull/279 and https://github.com/piraeusdatastore/piraeus-operator/pull/282 merged first

This PR introduces:

- `--bind-address` option for piraeus-operator to specify bind address for operator metrics exporter

- `csi.controllerSidecars`, `csi.nodeSidecars` `operator.sidecars`, `operator.controller.sidecars`, `operator.satelliteSet.sidecars` helm options to have an ability specify additional containers for pods
- `csi.controllerExtraVolumes`, `csi.nodeExtraVolumes`, `operator.extraVolumes`, `operator.controller.extraVolumes` ,  `operator.satelliteSet.extraVolumes` helm options to have an ability specify additional volumes for pods
- `operator.args.bindAddress`, `operator.controller.httpBindAddress`, `operator.controller.httspBindAddress`, `operator.satelliteSet.monitoringBindAddress` to have an ability specify bind address for metrics collectors and so on

adds example values and resources to configure rbac-proxy and podMonitors for prometheus-operator


pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -1148,9 +1149,9 @@ func newMonitoringConfigMap(set *piraeusv1.LinstorSatelliteSet) *corev1.ConfigMa
 		Data: map[string]string{
 			"prometheus.toml": fmt.Sprintf(`
 [[prometheus]]
-address = "0.0.0.0:%d"
+address = "%s:%d"
 enums = true
-`, monitoringPort),
+`, set.Spec.MonitoringBindAddress, monitoringPort),
```
WanzenBug, line 1154: I think we need to handle the "empty" case here, since this is a new field that might not be set. Configuring drbd-reactor with `address = ":9942"` does not work. interestingly, it doesn't produce any errors, but it also is not reachable from anywhere.

WanzenBug (changes requested):
Looks good. One small issue with compatibility, and this is needs a proper entry in CHANGELOG.md, and maybe also a note in UPGRADE.md.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
kvaps: Thanks, fixed

examples/rbac-proxy-resources.yaml
```diff
@@ -0,0 +1,184 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: piraeus-rbac-proxy
+data:
+  piraeus-operator.yaml: |+
+    authorization:
+      resourceAttributes:
+        namespace: linstor
```
WanzenBug, line 10: I'd replace the example namespace with `piraeus`. Also add a comment at the start of the file about the assumed namespaces

examples/rbac-proxy-resources.yaml
```diff
@@ -0,0 +1,184 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: access-to-linstor
+subjects:
+- kind: User
+  name: d8-monitoring:scraper
+- kind: ServiceAccount
+  name: prometheus
+  namespace: d8-monitoring
```
WanzenBug, line 184: Basically same as above here. kube-prometheus "normal" destination is the `monitoring` namespace so we should use that here and mention it in a comment at the start.

WanzenBug (changes requested):
Was testing it a bit and noticed some small issues.

examples/rbac-proxy-values.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - name: rbac-proxy-config
+      configMap:
+        name: piraeus-rbac-proxy
+  controller:
+    httpBindAddress: "127.0.0.1"
+    sidecars:
+      - name: kube-rbac-proxy
+        image: quay.io/brancz/kube-rbac-proxy:v0.11.0
+        args:
+          - "--upstream=http://127.0.0.1:3370"
```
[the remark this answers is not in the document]
kvaps: LINSTOR serves on both ports 3370 (http) and 3371 (https), however we switch http to listen only 127.0.0.1 to protect the metrics endpoint:

https://github.com/piraeusdatastore/piraeus-operator/blob/d1677edd97cb3b66a2bb2b9f84cdfbba68d7c4e2/examples/rbac-proxy-values.yaml#L32

while rbac-proxy simple consumes it

examples/rbac-proxy-resources.yaml
kvaps: fixed, thanks

examples/rbac-proxy-resources.yaml
kvaps: fixed, thanks

PR #281, opened 2022-03-01
controller: avoid error loop when scaling deployment

WanzenBug:
We might run into situations where we need to create a backup of LINSTOR DB
resources while no LINSTOR Controller deployment is active. For example,
this may happen when a user temporarily uninstalls the CR, or similar.

While our code to determine the image version deals with this gracefully,
we forgot about this possibility when scaling the deployment. If we
can't find the deployment to patch, that already implies that there
is nothing for us to stop, so we can continue without erroring out.

See https://github.com/piraeusdatastore/piraeus-operator/pull/246#issuecomment-1055467169


kvaps (approved):
Working fine 👍

WanzenBug:
@JoelColledge please review

JoelColledge (approved):
Sorry, missed the notification for this one. Looks good.

PR #282, opened 2022-03-03
Introduce option to disable monitoring resources management

kvaps:
Current implementation of serviceMonitors is not flexible.

It has few potential problems:
- Quite often monitoring systems require setting of additional labels for `ServiceMonitors` and run them in different namespaces (exactly our case)
- Using `ServiceMonitor` makes not much sense for piraeus-operator and linstor-node daemonset, `PodMonitor` works as well but does not require any additional `Service` for this
- operator-sdk v0.19 has [hardcoded label](https://github.com/operator-framework/operator-sdk/blob/125d0dfcc71fef4f9d7e2a42b1354cb79ffdee03/pkg/metrics/metrics.go#L112) `name` which must be set for both `Service` and `Pod`, otherwise it will not work, but your operator can be running with the different set of labels, especially when you follow [Recommended Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) guide of Kubernetes
- All thesee fancy api are not supported since operator-sdk v1.0.0

Given all of the above, I have decided not to try to solving all the problems, but do that on another level. Namely to provide user an oportunity to define his own monitoring resources.

This PR introduces new option `--create-monitoring=<bool>`, which is enabled by default. Later it also can be considered to move monitroing configuration into Helm chart and deprecate the option.

### DIFF
--- a/pkg/controller/linstorcontroller/linstorcontroller_controller.go
+++ b/pkg/controller/linstorcontroller/linstorcontroller_controller.go
@@ -56,6 +56,9 @@ import (
 // CreateBackups controls if the operator will create a backup of the LINSTOR resources before upgrading.
 var CreateBackups = true
 
+// CreateMonitoring controls if the operator will create a monitoring resources.
+var CreateMonitoring = true
+
 // newControllerReconciler returns a new reconcile.Reconciler
 func newControllerReconciler(mgr manager.Manager) reconcile.Reconciler {
 	return &ReconcileLinstorController{client: mgr.GetClient(), scheme: mgr.GetScheme()}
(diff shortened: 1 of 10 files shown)


pkg/controller/linstorcontroller/linstorcontroller_controller.go
```diff
@@ -56,6 +56,9 @@ import (
 // CreateBackups controls if the operator will create a backup of the LINSTOR resources before upgrading.
 var CreateBackups = true
 
+// CreateMonitoring controls if the operator will create a monitoring resources
+var CreateMonitroing = true
```
WanzenBug, line 60:
```suggestion
var CreateMonitoring = true
```

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -56,6 +57,9 @@ import (
 	lc "github.com/piraeusdatastore/piraeus-operator/pkg/linstor/client"
 )
 
+// CreateMonitoring controls if the operator will create a monitoring resources
+var CreateMonitroing = true
```
WanzenBug, line 61:
```suggestion
var CreateMonitoring = true
```

WanzenBug (changes requested):
A few typos. Also, not sure what happened with the formatting in `cmd/manager.main.go`

WanzenBug (approved):

PR #290, opened 2022-03-25
reconfigure net-interface if transport changed

WanzenBug:
If a user adds the SSL certificates to an existing cluster the existing
connection needs to be reconfigured to use the SSL type and port. A naive
check for the satellites address is no longer enough.


JoelColledge (approved):

WanzenBug:
Ended up having to work around a inconsistency between the golinstor constants and what LINSTOR actually reports. Itseems any change to the (generated) constants had implications to LINSTOR internals, so I've just changed the comparison to be case-insensitive.

JoelColledge (approved):

PR #295, opened 2022-04-08
chart: add toleration for control plane

WanzenBug:
In Kubernetes 1.24, the master roles is renamed to control-plane, and all
node taints are renamed. To keep existing behaviour, include tolerations
for both taints.


kvaps (approved):
lgtm

PR #298, opened 2022-04-14
satellites: restore evicted satellites if pod exists

### LINKED ISSUE #297: Automatic restore of evicted nodes
If node come offline for 2 hours then return back it stays `EVICTED` even when linstor-satellite succefully started.
Shouldn't we recover such nodes automatically by piraeus-operrator?

WanzenBug:
Restore satellites for those pods we know are running. Even if the satellite
is still offline, the initial eviction should have already triggered all the
replacement resource creations.

Closes #297

### DIFF
--- a/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
+++ b/pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
@@ -447,6 +447,12 @@ func (r *ReconcileLinstorSatelliteSet) reconcileAllNodesOnController(ctx context
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


JoelColledge (approved):

kvaps:
It does not work to me for some reason. What I do:

```bash
kubectl edit nodes.internal.linstor.linbit.com <node>
```
set `node_flags: 3` for it, then restart `linstor-controller`

node remains `EVICTED` even after restart `linstor-node` and `piraeus-operator` pods

WanzenBug:
> It does not work to me for some reason

Just tested it again, and it definitly does work for me, using both "real" evictions and the DB hack you used.

PR #300, opened 2022-04-26
Remove backup nagging

WanzenBug:
Operator 1.8.0 already creates a backup, so explicit confirmation by the
user is no longer needed. Instead, add some warnings in the documentation
for users upgrading from before 1.8.0.


JoelColledge (approved):
Looks good to me.

PR #302, opened 2022-05-05
Update Host Setup guide

Dr-Shadow:
Host setup guide is missing a new dependency on libelf and a link to the drbd9-jammy image.

I added a Dockerfile for Ubuntu 22.04 as well since the dependencies are named differently.

### DIFF
--- a/doc/host-setup.md
+++ b/doc/host-setup.md
@@ -39,10 +39,13 @@ for the kernel source:
 
 Install these packages and use the injector image matching your distribution:
 
-* CentOS 7: `quay.io/piraeusdatastore/drbd9-centos7`
-* CentOS 8: `quay.io/piraeusdatastore/drbd9-centos8` (See [#137] for CentOS 8 Stream)
-* Ubuntu 20.04: `quay.io/piraeusdatastore/drbd9-focal`
-* Ubuntu 18.04: `quay.io/piraeusdatastore/drbd9-bionic`
+| Distribution | Docker image                                                              | Original Dockerfile                                                                                                             |
+|--------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


doc/host-setup.md
```diff
@@ -111,6 +112,28 @@ ENTRYPOINT ["sh", "-e", "/entry-override.sh"]
 Build the image, tag it and push it to a registry accessible to your cluster. You can start using your custom injector
 image by passing `--set operator.satelliteSet.kernelModuleInjectionImage=<your-custom-image>` to Helm.
 
+For Ubuntu 22.04, we supply a `Dockerfile` as an example as well
```
WanzenBug, line 115: Why would you need this file? I mean what is different from https://github.com/piraeusdatastore/piraeus/blob/master/dockerfiles/drbd-driver-loader/Dockerfile.jammy ?

WanzenBug (changes requested):
Please ensure https://github.com/piraeusdatastore/piraeus-operator/pull/302/checks passes. i.e. you need to `--signoff` your commit and agree to the CNCF terms.

Also not sure what the jammy docker file adds that the existing file doesn't cover?

Dr-Shadow:
> Please ensure https://github.com/piraeusdatastore/piraeus-operator/pull/302/checks passes. i.e. you need to `--signoff` your commit and agree to the CNCF terms.

I will do

> Also not sure what the jammy docker file adds that the existing file doesn't cover?

Oh you're right, should I just make a link to this file ?

I had to make custom images for Hirsute (coccinelle was missing for example) so I already had a custom image without an overriding entry. It just broke with the new elfutils dependency when I upgraded to jammy.

WanzenBug:
> Oh you're right, should I just make a link to this file ?

I think you should just link the dockerfiles from https://github.com/piraeusdatastore/piraeus/tree/master/dockerfiles/drbd-driver-loader in the initial list of docker images, and remove the added example.

If someone needs to build for other ubuntu versions, they should be able to figure it out from the linked docker files

doc/host-setup.md
Dr-Shadow: I just added the suggested modifications

WanzenBug (approved):

PR #303, opened 2022-05-09
update dependencies

WanzenBug:
Update all dependencies. Most notabely, this removes code dependency on
operator-sdk. There were also reported crashes fixed in newer versions of
k8s.io/client-go and others.

One change this does introduce is the removal of prometheus operator resources
for the operator itself. These were just the default metrics on the CR, and as
such not that interesting.


cmd/manager/main.go
```diff
@@ -18,214 +18,83 @@ limitations under the License.
...
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		Namespace:              os.Getenv("WATCH_NAMESPACE"),
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "363fd0da.piraeus.io",
```
JoelColledge, line 73: Where does this magic number come from?

go.mod
```diff
@@ -1,22 +1,89 @@
...
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	k8s.io/apiextensions-apiserver v0.24.0 // indirect
+	k8s.io/component-base v0.24.0 // indirect
+	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20220413171646-5e7f5fdc6da6 // indirect
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
```
JoelColledge, line 88: Why do we have so many indirect dependencies?

JoelColledge (commented):
Great. I would have expected removing the operator-sdk code dependency to be more disruptive.

cmd/manager/main.go
WanzenBug: Oops, I let `operator-sdk init` generate a new temporary project that I used to lift the new "main" code... looks like it automatically generated this ID.

go.mod
WanzenBug: I think this is because I updated using go1.18. Looking at https://go.dev/ref/mod#graph-pruning to me it reads like requiring a module with requirement <1.17 automatically includes all those dependencies as indirect. Since a lot of the dependencies are not >go1.17 we end up with a lot of transitive dependencies.

WanzenBug: @JoelColledge anything I should investigate further?

JoelColledge: Note from my side. Looks like this is just how it is for now.

JoelColledge (approved):

Issue #270, opened 2022-02-08
Scheduling extender plugin for LINSTOR


kvaps:
Follows up https://github.com/piraeusdatastore/piraeus-operator/pull/219#issuecomment-1027282815

Yeah, now I understand why you decided to get rid of stork. It's quite huge, slow, contains a lot of complex controllers and requires ultimate permissions on cluster and does not support secure communication with kube-scheduller.
As a result, it is slowly updated.

Nevertheless, I found that stork [scheduler extender](https://github.com/libopenstorage/stork/tree/master/pkg/extender) component does not require such maintenance.

Eg. it is working fine with Kubernetes ±1.22. Only config format for kube-scheduler was [changed](https://github.com/libopenstorage/stork/pull/982).

I'm still convinced in my point. We shouldn't use Kubernetes CSI topology feature until it can be dynamically updated after volume is created. So I made a decision to write separate linstor scheduler extender based on existing stork driver, and run it as sidecar to the kube-scheduler pod.

By this I solved following problems:
- Now we have independed linstor-scheduller with small amount of code and superfluous controllers/CRDs/whatever
- Permissions are reduced to same amount as Kube-sheduler have.
- Secure communication over localhost instead of sending raw HTTP requests over Kubernetes networking.
- The code is backward-compatible with current stork-driver.

Link to the project: https://github.com/kvaps/linstor-scheduler-extender
Take a look at deploy folder, it has self-sufficient example.

WanzenBug:
Nice work. I'll try to take a closer look at the project soon. One of the things I'd like to make sure is that this project does:

* Work with topology, if enabled. I don't want to lock people into using only topology or only the scheduler plugin.
* Work with late binding. This is something that broke STORK when I last tried it. I.e. the scheduler gets called for a Pod which has an unbound volume claim, because the claim is waiting for the volume to be scheduled first.

I haven't looked at what is reused from the STORK plugin, but if it's not too much work I would like to not depend on Stork. That way, it is much easier to update.

kvaps:
> Work with late binding. This is something that broke STORK when I last tried it. I.e. the scheduler gets called for a Pod which has an unbound volume claim, because the claim is waiting for the volume to be scheduled first.

Works as a charm. Linstor-scheduler just provides similar ranks for all the nodes for the first binding, letting Kubernetes to decide.

> Work with topology, if enabled. I don't want to lock people into using only topology or only the scheduler plugin.

[...]

PR #307, opened 2022-05-12
Inconsistency in stork configmap volume name

vholer:
Installation with Stork enabled fails as the config volume name is different to what is used in container mount reference:

https://github.com/piraeusdatastore/piraeus-operator/blob/e47426de7f36305073a4c84dfb2f054d5630cfb3/charts/piraeus/templates/stork-deployment.yaml#L301-L303

Error output from helm:

```
Release "piraeus" does not exist. Installing it now.
Error: Deployment.apps "piraeus-stork-scheduler" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "stork-config"
```


vholer:
Thanks, @kvaps. Yep, I have seen some ongoing communication in #270.

kvaps (approved):
LGTM

PR #311, opened 2022-05-24
chart: fix ugprading from old CRDs

WanzenBug:
Using `httpBindAddress` caused validation to fail if upgrading from older
releases with TLS configured. Since LINSTOR already redirects from HTTP
to HTTPS in all cases, the default http address does not need to be
reconfigured when TLS is used.

### DIFF
--- a/charts/piraeus/templates/operator-controller.yaml
+++ b/charts/piraeus/templates/operator-controller.yaml
@@ -38,8 +38,6 @@ spec:
   {{- end }}
   {{- if .Values.operator.controller.httpBindAddress }}
   httpBindAddress: {{ .Values.operator.controller.httpBindAddress | quote }}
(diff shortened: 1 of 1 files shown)


WanzenBug:
> What about /metrics?

You are right, it seems like LINSTOR doesn't redirect in that case. I think that is actually good for rbac-proxy, which would otherwise fail to connect to LINSTOR I think. In the example you provided we explicitly set the `httpBindAddress`, so we don't need this fallback.

Admittedly, this is not 100% ideal, as it means the metrics are always unprotected.

kvaps:
Ah, got it, I'll just leave links to original issues:
- https://github.com/helm/helm/issues/8668
- https://github.com/helm/helm/pull/6243
- https://github.com/helm/helm/issues/10784

kvaps (approved):
LGTM

Issue #313, opened 2022-06-06
Support custom labels and annotations via values.yaml


bc185174:
Support the application of custom labels and annotations through the `values.yaml`. Would be nice for some granularity over these such that it can be adjusted for each component, i.e.

```yaml
csi:
  controllerLabels:
    test: "example"
  controllerAnnotations:
    test: "example"
  nodeLabels: {}
  nodeAnnotations: {}
operator:
  customLabels: {}
  customAnnotations: {}
  controller:
    customLabels: {}
    customAnnotations: {}
  satelliteSet:
    customLabels: {}
    customAnnotations: {}
haController:
  customLabels: {}
  customAnnotations: {}
```

WanzenBug:
Implemented in #314

PR #314, opened 2022-06-07
Support custom labels and annotations

bc185174:
Values file update to allow custom labels and annotations:

```yaml
customLabels: {}
customAnnotations: {}
```

These can be applied to the following components:

- HA Controller
- CSI Components
- Satellite Sets
- Linstor Controller
- Operator

Issue #313

### DIFF
--- a/charts/piraeus/templates/ha-controller-deployment.yaml
+++ b/charts/piraeus/templates/ha-controller-deployment.yaml
@@ -4,6 +4,19 @@ kind: Deployment
 metadata:
   name: {{ template "operator.fullname" . }}-ha-controller
   namespace: {{ .Release.Namespace }}
+{{- if .Values.haController.customAnnotations }}
+  annotations:
+{{- with .Values.haController.customAnnotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
+  labels:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-ha-controller
+{{- if .Values.haController.customLabels }}
+{{- with .Values.haController.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
 spec:
   replicas: {{ .Values.haController.replicas }}
   selector:
@@ -13,6 +26,17 @@ spec:
     metadata:
       labels:
         app.kubernetes.io/name: {{ template "operator.fullname" . }}-ha-controller
+{{- if .Values.haController.customLabels }}
+{{- with .Values.haController.customLabels }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- end }}
+{{- if .Values.haController.customAnnotations }}
+      annotations:
+{{- with .Values.haController.customAnnotations }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- end }}
     spec:
       serviceAccountName: ha-controller
       initContainers:
--- a/charts/piraeus/templates/operator-controller.yaml
+++ b/charts/piraeus/templates/operator-controller.yaml
@@ -4,6 +4,19 @@ kind: LinstorController
 metadata:
   name: {{ template "operator.fullname" . }}-cs
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-cs
+{{- if .Values.operator.controller.customLabels }}
+{{- with .Values.operator.controller.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
+{{- if .Values.operator.controller.customAnnotations }}
+  annotations:
+{{- with .Values.operator.controller.customAnnotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
 spec:
   priorityClassName: {{ .Values.priorityClassName | default "" | quote }}
   # TODO: switch to k8s db by default
--- a/charts/piraeus/templates/operator-csi-driver.yaml
+++ b/charts/piraeus/templates/operator-csi-driver.yaml
@@ -4,6 +4,19 @@ kind: LinstorCSIDriver
 metadata:
   name: {{ template "operator.fullname" . }}
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}
+{{- if .Values.csi.customLabels }}
+{{- with .Values.csi.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
+{{- if .Values.csi.customAnnotations }}
+  annotations:
+{{- with .Values.csi.customAnnotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
 spec:
   imagePullSecret: {{ .Values.drbdRepoCred | quote }}
   linstorPluginImage: {{ .Values.csi.pluginImage | quote }}
--- a/charts/piraeus/templates/operator-deployment.yaml
+++ b/charts/piraeus/templates/operator-deployment.yaml
@@ -5,6 +5,17 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     name: {{ template "operator.fullname" . }}-operator
+{{- if .Values.operator.customLabels }}
+{{- with .Values.operator.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
+{{- if .Values.operator.customAnnotations }}
+  annotations:
+{{- with .Values.operator.customAnnotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
 spec:
   replicas: {{ .Values.operator.replicas }}
   selector:
@@ -14,6 +25,17 @@ spec:
     metadata:
       labels:
         name: {{ template "operator.fullname" . }}-operator
+{{- if .Values.operator.customLabels }}
+{{- with .Values.operator.customLabels }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- end }}
+{{- if .Values.operator.customAnnotations }}
+      annotations:
+{{- with .Values.operator.customAnnotations }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- end }}
     spec:
       serviceAccountName: {{ template "operator.fullname" . }}
       securityContext:
--- a/charts/piraeus/templates/operator-satelliteset.yaml
+++ b/charts/piraeus/templates/operator-satelliteset.yaml
@@ -4,6 +4,19 @@ kind: LinstorSatelliteSet
 metadata:
   name: {{ template "operator.fullname" . }}-ns
   namespace: {{ .Release.Namespace }}
+  labels:
+    name: {{ template "operator.fullname" . }}-ns
+{{- if .Values.operator.satelliteSet.customLabels }}
+{{- with .Values.operator.satelliteSet.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+{{- end }}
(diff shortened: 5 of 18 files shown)


charts/piraeus/templates/ha-controller-deployment.yaml
```diff
@@ -4,6 +4,16 @@ kind: Deployment
 metadata:
   name: {{ template "operator.fullname" . }}-ha-controller
   namespace: {{ .Release.Namespace }}
+  annotations:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-ha-controller
```
WanzenBug, line 8: I don't think this needs to be an annotation, just the label is enough.

charts/piraeus/templates/ha-controller-deployment.yaml
```diff
@@ -13,6 +23,14 @@ spec:
     metadata:
       labels:
         app.kubernetes.io/name: {{ template "operator.fullname" . }}-ha-controller
+{{- with .Values.haController.customLabels }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+      annotations:
+        app.kubernetes.io/name: {{ template "operator.fullname" . }}-ha-controller
```
WanzenBug, line 30: same as before: label should be good enough

charts/piraeus/templates/operator-controller.yaml
```diff
@@ -4,6 +4,16 @@ kind: LinstorController
 metadata:
   name: {{ template "operator.fullname" . }}-cs
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-operator-controller
```
WanzenBug, line 8: should be the same as the `name` field I think, so `{{ template "operator.fullname" . }}-cs`

charts/piraeus/templates/operator-controller.yaml
```diff
@@ -4,6 +4,16 @@ kind: LinstorController
...
+{{- with .Values.operator.controller.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+  annotations:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-operator-controller
```
WanzenBug, line 13: again, remove from annotations.

charts/piraeus/templates/operator-csi-driver.yaml
```diff
@@ -4,6 +4,16 @@ kind: LinstorCSIDriver
 metadata:
   name: {{ template "operator.fullname" . }}
   namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-csi-driver
+{{- with .Values.csi.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+  annotations:
+    app.kubernetes.io/name: {{ template "operator.fullname" . }}-csi-driver
```
WanzenBug, line 13: same as before. should be matching the resource name and remove the name from annotations.

(Yes, I know our naming is a bit messed up and not perfectly logical, but I think having it consistent is better than only fixing it half-way)

charts/piraeus/templates/operator-deployment.yaml
```diff
@@ -5,6 +5,14 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     name: {{ template "operator.fullname" . }}-operator
+{{- with .Values.operator.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+  annotations:
+    name: {{ template "operator.fullname" . }}-operator
```
WanzenBug, line 12: and one more

charts/piraeus/templates/operator-deployment.yaml
```diff
@@ -14,6 +22,14 @@ spec:
     metadata:
       labels:
         name: {{ template "operator.fullname" . }}-operator
+{{- with .Values.operator.customLabels }}
+{{ toYaml . | indent 8 }}
+{{- end }}
+      annotations:
+        name: {{ template "operator.fullname" . }}-operator
```
WanzenBug, line 29: and another

charts/piraeus/templates/operator-satelliteset.yaml
```diff
@@ -4,6 +4,16 @@ kind: LinstorSatelliteSet
 metadata:
   name: {{ template "operator.fullname" . }}-ns
   namespace: {{ .Release.Namespace }}
+  labels:
+    name: {{ template "operator.fullname" . }}-satellite-set
+{{- with .Values.operator.satelliteSet.customLabels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+  annotations:
+    name: {{ template "operator.fullname" . }}-satellite-set
```
WanzenBug, line 13: and here again: use the same name as in the `name` field, and remove from annotations

pkg/controller/linstorcontroller/linstorcontroller_controller.go
```diff
@@ -1056,12 +1056,24 @@ func expectedEndpoint(controllerResource *piraeusv1.LinstorController) string {
...
+		Annotations: controllerResource.ObjectMeta.Annotations,
+	}
+}
+
+func getControllerLabels(controllerResource *piraeusv1.LinstorController) map[string]string {
+	return mergeStringMap(map[string]string{
+		"app.kubernetes.io/name":       kubeSpec.ControllerRole,
+		"app.kubernetes.io/instance":   controllerResource.Name,
+		"app.kubernetes.io/managed-by": kubeSpec.Name,
+	}, controllerResource.ObjectMeta.Labels)
```
WanzenBug, line 1071: We have to be careful not to include these "user" labels in the `Selector:` for deployments. Otherwise we could get into a situation where a user modified a label, but the operator can't proceed, because updating the selector value isn't possible.

pkg/controller/linstorcontroller/linstorcontroller_controller.go
```diff
@@ -1056,12 +1056,24 @@ func expectedEndpoint(controllerResource *piraeusv1.LinstorController) string {
...
+}
+
+func mergeStringMap(stringMapA map[string]string, stringMapB map[string]string) map[string]string {
```
WanzenBug, line 1074: Move this and the copies in the other controllers to `pkg/k8s/metadata/util/util.go`

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -1054,17 +1054,29 @@ func defaultIfUnset(val, def int32) int32 {
...
 	return filepath.Join(append([]string{csiResource.Spec.KubeletPath}, subdirs...)...)
 }
+
+func getCSILabels(controllerResource *piraeusv1.LinstorCSIDriver, component string) map[string]string {
+	return mergeStringMap(map[string]string{
+		"app.kubernetes.io/name":       kubeSpec.CSIDriverRole,
+		"app.kubernetes.io/instance":   controllerResource.Name,
+		"app.kubernetes.io/managed-by": kubeSpec.Name,
+		"app.kubernetes.io/component":  component,
+	}, controllerResource.ObjectMeta.Labels)
```
WanzenBug, line 1074: Same applies to this: we don't want the user controlled labels in the `Selector`

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -1509,14 +1509,26 @@ func findK8sNode(nodes []corev1.Node, name string) *corev1.Node {
...
+		Annotations: satelliteSet.Annotations,
+	}
+}
+
+func getSatelliteLabels(satelliteSet *piraeusv1.LinstorSatelliteSet) map[string]string {
+	return mergeStringMap(map[string]string{
+		"app.kubernetes.io/name":       kubeSpec.NodeRole,
+		"app.kubernetes.io/instance":   satelliteSet.Name,
+		"app.kubernetes.io/managed-by": kubeSpec.Name,
+	}, satelliteSet.ObjectMeta.Labels)
```
WanzenBug, line 1524: And here too: no user controlled labels in the `Selector:`

WanzenBug (changes requested):
Changes look good in principal. There are a few issues regarding some default names and annotations, as well as the selectors for deployment/daemonset.

Please also add a short entry in the changelog.

bc185174:
The Selector fields should only be using the default labels now. The custom ones can only be applied to the metadata.

WanzenBug (changes requested):
Please also run `make helm-values deploy/piraeus`, which should regenerate some files.

WanzenBug (approved):
I'll do some more internal tests, but generally LGTM. I'll also squash all commits into one, no need for 14 new commits for this feature.

pkg/controller/linstorcontroller/linstorcontroller_controller.go
```diff
@@ -1054,14 +1054,20 @@ func expectedEndpoint(controllerResource *piraeusv1.LinstorController) string {
...
-		Name:      fmt.Sprintf(nameFmt, controllerResource.Name),
-		Namespace: controllerResource.Namespace,
-		Labels: map[string]string{
-			"app.kubernetes.io/name":       kubeSpec.ControllerRole,
-			"app.kubernetes.io/instance":   controllerResource.Name,
-			"app.kubernetes.io/managed-by": kubeSpec.Name,
-		},
+		Name:        fmt.Sprintf(nameFmt, controllerResource.Name),
+		Namespace:   controllerResource.Namespace,
+		Labels:      mdutil.MergeStringMap(defaultLabels, controllerResource.ObjectMeta.Labels),
```
WanzenBug, line 1070: The order of arguments needs to be reversed here. Otherwise the custom labels overwrite our defaultLabels, and then the selector does not match. This is then rejected by k8s, so the resource can never start.

pkg/controller/linstorcsidriver/linstorcsidriver_controller.go
```diff
@@ -1052,19 +1052,25 @@ func defaultIfUnset(val, def int32) int32 {
...
-		Namespace: controllerResource.Namespace,
-		Labels: map[string]string{
-			"app.kubernetes.io/name":       kubeSpec.CSIDriverRole,
-			"app.kubernetes.io/instance":   controllerResource.Name,
-			"app.kubernetes.io/managed-by": kubeSpec.Name,
-			"app.kubernetes.io/component":  component,
-		},
+		Name:        fmt.Sprintf(nameFmt, controllerResource.Name),
+		Namespace:   controllerResource.Namespace,
+		Labels:      mdutil.MergeStringMap(defaultLabels, controllerResource.ObjectMeta.Labels),
```
WanzenBug, line 1060: Same issue here.

pkg/controller/linstorsatelliteset/linstorsatelliteset_controller.go
```diff
@@ -1508,14 +1508,20 @@ func findK8sNode(nodes []corev1.Node, name string) *corev1.Node {
...
-		Name:      fmt.Sprintf(nameFmt, satelliteSet.Name),
-		Namespace: satelliteSet.Namespace,
-		Labels: map[string]string{
-			"app.kubernetes.io/name":       kubeSpec.NodeRole,
-			"app.kubernetes.io/instance":   satelliteSet.Name,
-			"app.kubernetes.io/managed-by": kubeSpec.Name,
-		},
+		Name:        fmt.Sprintf(nameFmt, satelliteSet.Name),
+		Namespace:   satelliteSet.Namespace,
+		Labels:      mdutil.MergeStringMap(defaultLabels, satelliteSet.ObjectMeta.Labels),
```
WanzenBug, line 1515: And here.

WanzenBug (changes requested):

WanzenBug (dismissed):

WanzenBug (approved):

WanzenBug (approved):

WanzenBug:
Thank you for your contribution!

Issue #318, opened 2022-06-21
Add option to deploy with Kustomize


bc185174:
Looking to add a Kustomization manifest to deploy the piraeus operator. The `deploy` directory would be a good `base` that others can patch/override.

PR #319, opened 2022-06-21
add option to deploy piraeus via kustomize

bc185174:
Provides base kustomization manifest to deploy the operator.


bc185174:
#318

WanzenBug (changes requested):
In general let's get this merged, but I'd like to see some automation for updating `kustomization.yaml`. There is already a makefile that generates the whole `deploy/piraeus` directory, so we should be able to generate the file list, too.

bc185174:
> In general let's get this merged, but I'd like to see some automation for updating `kustomization.yaml`. There is already a makefile that generates the whole `deploy/piraeus` directory, so we should be able to generate the file list, too.

Yeah its a little tricky to generate but I'll see what I can do to automate it.

deploy/create-kustomization.bash
```diff
@@ -0,0 +1,24 @@
...
+resources:
+EOL
+
+
+files="$(find -L "piraeus" -type f)"
+echo "$files" | while read -r file; do
+  echo "  - $file" >> "$kustomize"
+done
+
+popd
```
WanzenBug, line 24: Please add a new lineline at the end here.

deploy/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - piraeus/templates/operator-csi-driver.yaml
+  - piraeus/templates/operator-controller.yaml
+  - piraeus/templates/satellite-rbac.yml
+  - piraeus/templates/operator-deployment.yaml
+  - piraeus/templates/controller-rbac.yml
+  - piraeus/templates/ha-controller-rbac.yaml
+  - piraeus/templates/csi-node-rbac.yml
+  - piraeus/charts/etcd/templates/rbac.yaml
+  - piraeus/charts/etcd/templates/service.yaml
+  - piraeus/charts/etcd/templates/statefulset.yaml
```
WanzenBug, line 22: When I run `make deploy/piraeus`, I get changes to this file. I think the script needs to `find ... | sort`, and then regenerate this file, so we don't get random changes to this file

WanzenBug (changes requested):
Almost ready to merge, just a few small issues to iron out still,

WanzenBug (approved):

bc185174:
Added change log entry and updated the bash script to sort the files first.

WanzenBug (dismissed):

WanzenBug (approved):

Issue #321, opened 2022-07-20
Can't remove nodeAffinity


kvaps:
When unsetting `nodeAffinity` the deployment is not going to updated.

I tried both ways:

```yaml
affinity: {}
```

and

```yaml
affinity:
  nodeAffinity: {}
  podAntiAffinity: {}
```

checked with linstorcontroller, linstorsatelliteset and linstorcsiplugin resources

kvaps:
Isn't it a bug of strategicpatch?
https://github.com/kubernetes/kubernetes/issues/48164

kvaps:
After some digging I found out that affinity field does not get into `piraeus.linbit.com/last-applied-configuration` annotation.

PR #323, opened 2022-07-20
fix lastAppliedAnnotation

kvaps:
It seems the last applied annotation is never updated for existing objects.
This PR adds it to generate three way merge patch.

Some logic borrowed here:
https://github.com/kubernetes/kubectl/blob/b5fe0f6e9c65ea95a2118746b7e04822255d76c2/pkg/util/apply.go#L67-L117

fixes https://github.com/piraeusdatastore/piraeus-operator/issues/321


WanzenBug (dismissed):
I'm kind of amazed that nobody noticed this sooner. 

Also: I need to step up my copy&paste game :man_facepalming:

WanzenBug (approved):
Thanks again for spotting this issue. LGTM

PR #333, opened 2022-09-02
Fix controller's service monitor when using TLS

blampe:
The presence of `HttpsControllerSecret`, not `SslConfig`, is what enables the controller's HTTPS port and client secret.

https://github.com/piraeusdatastore/piraeus-operator/blob/master/pkg/controller/linstorcontroller/linstorcontroller_controller.go#L658

https://github.com/piraeusdatastore/piraeus-operator/blob/master/pkg/controller/linstorcontroller/linstorcontroller_controller.go#L787 

Prior to this fix, a controller with HTTPS secrets configured but _not_ SslConfig would generate a ServiceMonitor that would erroneously attempt to scrape over HTTP.

https://github.com/piraeusdatastore/piraeus-operator/blob/master/pkg/controller/linstorcontroller/linstorcontroller_controller.go#L242


WanzenBug (approved):
LGTM. Thanks for finding and fixing the issue.

PR #346, opened 2022-11-10
Use secure-port of stork-scheduler on liveness check when k8s version is higher than 1.22

Suckzoo:
`--port` flag, whose default is 10251, is completely deprecated since 1.22. After 1.22+, kube-scheduler only accepts `--secure-port` whose default is 10259.


WanzenBug (approved):

WanzenBug:
Thanks!

PR #349, opened 2022-11-14
Add default image configuration

WanzenBug:
Adds a subpackages to create kustomize overrides for container images, based on a simple file-based default configuration.

The configuration sets a specific repository base, and configures the actual image names + tags for all components in the pkg/resources package (with the exception of upstream k8s components).

It also supports passing in the ".status.nodeInfo.osImage" information from the k8s node object. This field seems to be populated by /etc/os-release's PRETTY_NAME, which we can use to automatically select the right DRBD image to load.


config/manager/default_images.yaml
```diff
@@ -0,0 +1,33 @@
...
+    image: piraeus-server
+  linstor-csi:
+    tag: v0.21.0
+    image: piraeus-csi
+  drbd-reactor:
+    tag: v0.9.0
+    image: drbd-reactor
+  drbd-module-loader:
+    tag: v9.2.1
+    image: drbd9-jammy # Fall back image: chose a very recent-ish kernel, which hopefully can compile whatever config is actually in use
```
JoelColledge, line 18: "very recent-ish" is a bit confusing. "Fallback" is one word. Maybe:

```suggestion
    image: drbd9-jammy # Fallback image: chose a fairly recent kernel, which can hopefully compile whatever config is actually in use
```

JoelColledge (commented):
Looks good. I'm starting to learn how Kustomize works.

PR #350, opened 2022-11-14
Add various utility functions

WanzenBug:
This is a collection of utilities that proved useful during development. In order to support some of them, we already add some API structs, but only subtypes for now.

The utils are:
* `AnyError`, which returns any errors that may happened during reconciliation
* `fieldpath`, which is directly imported from `k8s.io/kubernetes/pkg/fieldpath`, just so we don't have to import the complete k8s repo
* `jsonpatch`, just some types for validation
* Some kustomize patch wrapper for type-safe handling of patch creation + validation
* `prune`, to prune no longer needed resources during reconciliation
* property helpers, helping to create LINSTOR properties based on k8s values.

For more information on each utility, check the commit message for each commit

### DIFF
--- a/api/v1/patch.go
+++ b/api/v1/patch.go
@@ -1,0 +1,95 @@
+package v1
+
+import (
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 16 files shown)


JoelColledge (commented):
Great! Been practicing some test-driven reviewing over here :slightly_smiling_face:

PR #351, opened 2022-11-16
Add podpatcher utility

WanzenBug:
In the case of satellites, the operator will have to deal with raw Pods instead of Deployments or Daemonsets. This is because we have some very particular requirements:

* We potentially have different Pod Specs per node, for example to use different DRBD loader images or include different Satellite configs.
* Satellites should have stable hostnames. This is because LINSTOR does not take to kindly to changed "uname"s for deployed resources. Since we no longer run in the host network by default, we inherit the Pod name as "kernel" hostname.

The issue with using raw Pods is that you can't just patch them like other resources: Pods are mostly immutable. This could be dealt with quite easily, if not for the fact that updates to images are allowed.

Updated images would mean that the user would see an unexpected container restart. In addition, the operator could not easily tell which version of the container was actually running, as that is not something we can easily get out of the Pod status.

So our solution is a custom patch function for Pods, that:

* Checks if the Pod already exists
  * If not, just apply the normal patch
* If the new Pod Spec uses the same images, try a normal patch
* If the new Pod Spec uses different images, or the patch could not be applied: delete and recreate the Pod.

### DIFF
--- a/pkg/podpatcher/podpatcher.go
+++ b/pkg/podpatcher/podpatcher.go
@@ -1,0 +1,86 @@
+package podpatcher
+
+import (
+	"context"
+	"encoding/json"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Patch patches a Pod.
+//
+// The differences to the normal client.Patch method are:
+// * We recreate the Pod if the patch failed to apply, as Pods generally can't be altered dynamically.
+// * We also recreate for image changes. Normally, these would be applied by kubelet eventually, but will cause a
+//   "restarted container" event, which many users will have alerts for. It also can't be nicely controlled by the
+//   operator. So we just have to make sure to detect any changes.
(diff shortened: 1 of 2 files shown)


JoelColledge:
An alternative might be to create a deployment for each satellite. I'm guessing you have considered this already?

JoelColledge (approved):

WanzenBug:
I did consider that (actually, a `DaemonSet` per node), but decided against it, because I still had issues with the host name. Checking again, I see that there actually is a way to force k8s to set the hostname in a stable way, other than fixing the hostname.

So in theory it would work, now I just need to think about if creating 40 different DaemonSets for 40 nodes is a good idea.

JoelColledge:
There's also the question: Why not just create the pods? It's not recommended for users, but we're an operator so maybe its fine. I guess it has an impact, for instance, why a satellite crashes. Do we want the operator to have to monitor and restart the pods in such situations?

JoelColledge:
> Ah, now I remember why the `spec.hostname` values was not viable: It only allows "a lowercase RFC 1123 label", i.e. not a FQDN, so we can't just take the k8s node name.

You mean one can set the pod's hostname to an FQDN by setting its name to that value, but not by overriding via the hostname field? :upside_down_face:

WanzenBug:
Exactly :facepalm:

PR #352, opened 2022-11-16
v2: add boilerplate + CI infrastructure

WanzenBug:


main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "59ff86a9.piraeus.io",
```
JoelColledge, line 104: What is the magic number for?

WanzenBug: Good question :thinking: 

It got autogenerated by `operator-sdk`, but I'm not sure why it is needed...

JoelColledge: Discussion seem familiar? https://github.com/piraeusdatastore/piraeus-operator/pull/303#discussion_r867815867

WanzenBug: I think I will just change it to something from `pkg/vars`.

JoelColledge (approved):
I assume that leaving all the explanation text in the `Makefile` is deliberate. Should help diffing with an updated template or so.

PR #353, opened 2022-11-16
add linstorhelper

WanzenBug:
Add helper functions to deal with the LINSTOR API. This includes:

* Creating nodes in LINSTOR and keeping their interfaces up to date.
* Creating a logging facade for the client using the operator-sdk logger.
* Setting and updating properties, ensuring only properties we previously managed are updated/deleted.


pkg/linstorhelper/properties.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/piraeusdatastore/piraeus-operator/v2/pkg/vars"
+)
+
+const (
+	LastApplyProperty = linstor.NamespcAuxiliary + "/" + vars.ApplyAnnotation
+	ManagedByProperty = linstor.NamespcAuxiliary + "/" + vars.ManagedByLabel
+)
+
+// GeneratePropertiesPatch returns the modification that need to be applied to update the current properties.
+func GeneratePropertiesPatch(current, expected map[string]string) *lclient.GenericPropsModify {
```
JoelColledge, line 19: "Patch" has quite a specific meaning in Kubernetes. This is, of course, similar, but it is not an actual Kubernetes patch. The name confused me when I first read it. Maybe use "Modify" like the linstor client? (Even though the grammar is a bit jarring.)

pkg/linstorhelper/properties.go
```diff
@@ -0,0 +1,75 @@
...
+		if _, ok := expected[k]; !ok {
+			result.DeleteProps = append(result.DeleteProps, k)
+		}
+	}
+
+	return result
+}
+
+// EnsureLastApplyProperty ensures the LastApplyProperty is up-to-date.
+func EnsureLastApplyProperty(props map[string]string) map[string]string {
```
JoelColledge, line 53:
```suggestion
func UpdateLastApplyProperty(props map[string]string) map[string]string {
```

"Ensure" means to me just "make sure that we have one".

pkg/linstorhelper/client.go
```diff
@@ -0,0 +1,113 @@
...
+
+	c, err := lapi.NewClient(append(options, lapi.BaseURL(u))...)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{*c}, nil
+}
+
+// EnsureNode ensures a node in LINSTOR matches the given node object.
+func (c *Client) EnsureNode(ctx context.Context, node lapi.Node) (*lapi.Node, error) {
```
JoelColledge, line 55: "Ensure" means to me just "make sure that we have one" and not that we perform updates.

pkg/linstorhelper/properties.go
WanzenBug: What about `GeneratePropertiesModification`? Or `MakePropertiesModification`? I think both would be rather readable.

pkg/linstorhelper/client.go
WanzenBug: It would be an `Upsert`, but I don't really like that word. In the older helpers it was `GetOrCreateNode`, which was IMO was just wrong...

`CreateOrUpdateNode`?

pkg/linstorhelper/properties.go
JoelColledge: Both sound good to me. Take your pick.

pkg/linstorhelper/client.go
JoelColledge: > `CreateOrUpdateNode`?

:+1:

pkg/linstorhelper/properties_test.go
```diff
@@ -0,0 +1,121 @@
...
+		t.Run(tcase.name, func(t *testing.T) {
+			t.Parallel()
+
+			diff := linstorhelper.MakePropertiesModification(tcase.current, tcase.expected)
+			assert.Equal(t, tcase.result, diff)
+		})
+	}
+}
+
+func TestEnsureLastApplyProperty(t *testing.T) {
```
JoelColledge, line 82: Rename test to match.

pkg/linstorhelper/client.go
```diff
@@ -0,0 +1,113 @@
...
+
+		newNode, err := c.Nodes.Get(ctx, node.Name)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get newly created node %s: %w", node.Name, err)
+		}
+
+		existingNode = newNode
+	}
+
+	patch := MakePropertiesModification(existingNode.Props, node.Props)
```
JoelColledge, line 77: Rename variable to match.

pkg/linstorhelper/properties_test.go
WanzenBug: Good catch!

JoelColledge (approved):

PR #355, opened 2022-11-17
v2: Satellite API

WanzenBug:
LinstorSatellite defines the desired state of a single LINSTOR satellite in
a cluster. K8s node and Satellite are linked by name, so every LinstorSatellite
resource needs a matching node object.
    
The resource itself allows setting up storage pools, LINSTOR properties
as well as customizing the default resources by using kustomize patches.
    
In addition to the resource, we also add a webhook to validate incoming
LinstorSatellite resource. In the webhook we check for common misconfigurations,
so that we can reject them immediately.

Also fixes a small issue in the image version loader


api/v1/linstorsatellite_types.go
```diff
@@ -0,0 +1,86 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+type ClusterReference struct {
+	Name string `json:"name,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster
```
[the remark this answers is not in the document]
WanzenBug: Yeah, they just got generated that way. Moved them "inline" now.

api/v1/storagepool.go
```diff
@@ -0,0 +1,268 @@
...
+		spNames.Insert(curSP.Name)
+		var oldSP *LinstorStoragePool
+		for j := range oldSPs {
+			if oldSPs[j].Name == curSP.Name {
+				oldSP = &oldSPs[j]
+				break
+			}
+		}
+
+		switch {
```
JoelColledge, line 123: This validates that at least one of `curSP.LvmThin` and `curSP.Lvm` are set, but not that at most one is.

WanzenBug: Good catch. Added a test to cover that scenario

JoelColledge (approved):

PR #357, opened 2022-11-21
satellites: add reconciler

WanzenBug:
Adds the main control loop for LinstorSatellite resources. It deploys the k8s resources, registers the satellite and maintains storage pools and network interfaces.

### DIFF
--- a/controllers/linstorsatellite_controller.go
+++ b/controllers/linstorsatellite_controller.go
@@ -1,0 +1,527 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


controllers/linstorsatellite_controller.go
```diff
@@ -0,0 +1,527 @@
...
+
+			p, err := lc.Nodes.GetStoragePool(ctx, lsatellite.Name, pool.Name, &lclient.ListOpts{Cached: &cached})
+			if err != nil {
+				return err
+			}
+
+			existingPool = &p
+		}
+
+		patch := linstorhelper.MakePropertiesModification(existingPool.Props, expectedProperties)
```
JoelColledge, line 408: We renamed `patch`.

JoelColledge (commented):
Looks good to me.

WanzenBug:
@JoelColledge looks like we decoded into the wrong struct during startup, so I added a small commit here.

PR #359, opened 2022-11-22
api: add LinstorSatelliteConfiguration resources

WanzenBug:
A LinstorSatelliteConfiguration is configuration for LinstorSatellites, that is applied to one or more (or even all) LinstorSatellites.

The to-be-added LinstorCluster controller will, for each node:
* Merge all LinstorSatelliteConfiguration resources with selectors matching the node.
* Create a LinstorSatellite object with the result of that merging process.

As such, the LinstorSatelliteConfiguration has almost the same structure as a LinstorSatellite, with the addition of the a NodeSelector field. It also has basically the same additional validation steps performed by the webhook.


WanzenBug:
Added the merging helper for satellite configs as that is basically the only interesting thing this resource does

JoelColledge (approved):
:+1:

PR #360, opened 2022-11-23
api: add LinstorCluster resources

WanzenBug:
A LinstorCluster resource is the single entity controlling creation of the whole LINSTOR cluster + CSI driver. From it, the operator will create the expected LinstorSatellite resources, as well as the LINSTOR Controller, CSI Controller and CSI Node DaemonSet.

The resource itself allows for:
* Changing the default repository used to pull images.
* Deploying the whole storage stack on only a subset of nodes.
* Setting cluster wide LINSTOR properties (== at controller level).
* Applying kustomize patches to any created k8s resource.


api/v1/linstorcluster_types.go
```diff
@@ -0,0 +1,83 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LinstorClusterSpec defines the desired state of LinstorCluster
+type LinstorClusterSpec struct {
+	// Repository used to pull workload images.
+	// +kubebuilder:validation:Optional
+	Repository string `json:"repository,omitempty"`
+
+	// NodeSelector selects for which nodes LinstorSatellite resources will be created.
```
JoelColledge, line 29:
```suggestion
	// NodeSelector selects which nodes LinstorSatellite resources will be created for.
```

Or:

```suggestion
	// NodeSelector selects the nodes for which LinstorSatellite resources will be created.
```

Or a bit less technical but perhaps more meaningful for the user:

```suggestion
	// NodeSelector selects the nodes on which LINSTOR Satellites will be deployed.
```

api/v1/linstorcluster_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// See https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +kubebuilder:validation:Optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Properties to apply on the cluster level.
+	//
+	// Use to create default settings for DRBD that should apply to all resources or configure some other cluster wide
```
JoelColledge, line 36:
```suggestion
	// Use to create default settings for DRBD that should apply to all resources or to configure some other cluster wide
```

api/v1/linstorcluster_types.go
WanzenBug: Thanks, chose the last one.

PR #361, opened 2022-11-23
cluster: add reconciler

WanzenBug:
Adds the main control loop for LinstorCluster resources. It deploy k8s resources, including LinstorSatellites for every node based on the available LinstorSatelliteConfigurations.

It also configures the LINSTOR controller by making modifictions to the controller properties based on default values and user customization.

### DIFF
--- a/controllers/linstorcluster_controller.go
+++ b/controllers/linstorcluster_controller.go
@@ -1,0 +1,534 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"sort"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	netwv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
+	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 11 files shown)


WanzenBug:
I think this is the last bit of Go code for now :smile:

controllers/linstorcluster_controller.go
```diff
@@ -0,0 +1,571 @@
...
+
+	k := &kusttypes.Kustomization{
+		Namespace: r.Namespace,
+		Labels:    r.kustomLabels(lcluster),
+		Resources: []string{"satellite-common"},
+		Images:    imgs,
+		Patches:   append(utils.MakeKustPatches(lcluster.Spec.Patches...), saPatch...),
+	}
+
+	return r.Kustomizer.Kustomize(k)
```
JoelColledge, line 363: These lines are basically repeated 3 times. Extract into separate function? Or do `GetVersions` and `pullSecretPatch` once and pass the values in?

WanzenBug: Done. This led down an interesting rabbit hole, because kustomize wanted to set the namespace on `LinstorSatellite` resources, which is fine for applying, but tripped up our resource pruning code, as it thought "namespaced satellite != cluster-scoped satellite" and immediately tried to remove the satellite after applying

controllers/linstorcluster_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/piraeusdatastore/piraeus-operator/v2/pkg/conditions"
+)
+
+var _ = Describe("LinstorCluster controller", func() {
+	const (
+		DefaultTimeout       = 30 * time.Second
+		DefaultCheckInterval = 5 * time.Second
+	)
+
+	Context("When creating an empty LinstorCluster", func() {
```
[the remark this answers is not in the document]
WanzenBug: This is actually more complicated than expected. I did not think about the `LinstorSatellite` reconciler that is also running. That is setting up a finalizer on the resources, so I have to think about that during test setup + tear down. Will take a bit longer to fix.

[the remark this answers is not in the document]
WanzenBug: I just noticed why my test cluster behaved so weirdly. I thought the envtest environment is basically a full K8s API, and it is, but I thought the "garbage collection" is included in that. But that seems to come from another k8s component that is not included.

So my test teardown waits for some things to be collected, but that never happens :disappointed:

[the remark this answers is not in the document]
JoelColledge: That's a pain. Have you got a solution?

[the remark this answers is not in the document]
WanzenBug: Nothing yet, but maybe I'll think of something over the weekend.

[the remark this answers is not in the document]
WanzenBug: :sweat_smile: Oops, I wrote a garbage collector just for our tests: https://github.com/piraeusdatastore/piraeus-operator/pull/361/commits/c482aa68f80ef1b72b00a5b64166661b83aaaec4

pkg/k8sgc/k8sgc.go
```diff
@@ -0,0 +1,132 @@
...
+		err := g.client.Get(ctx, client.ObjectKey{Name: owner.Name, Namespace: obj.GetNamespace()}, ownObj)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return false, err
+		}
+
+		if ownObj.GetUID() == owner.UID {
+			return false, err
```
JoelColledge, line 127:
```suggestion
			return false, nil
```

pkg/k8sgc/k8sgc.go
```diff
@@ -0,0 +1,132 @@
...
+		toWatch = append(toWatch, gvk)
+	}
+
+	return &gc{
+		client:  cl,
+		toWatch: toWatch,
+	}, nil
+}
+
+func (g *gc) Run(ctx context.Context) (bool, error) {
```
JoelColledge, line 70: This doesn't collect objects with an ownership loop, but I guess we don't care about that. Not sure it is even permitted.

JoelColledge (commented):
Fantastic! I love it, that you wrote a garbage collector and tests for it too.

pkg/k8sgc/k8sgc.go
WanzenBug: It kind of works, but only if one of the resources in the loop gets deleted first. Strictly speaking, it no longer is a cycle at that point. But I think that is the best we can do.

WanzenBug: It's completely redundant for our purpose, since we only have a tree for our resources.
```
LinstorCluster   -->   LinstorSatellite   -->   Pods, etc...
             \----->   Deployment, DaemonSet, etc...
```

JoelColledge (approved):

PR #362, opened 2022-11-30
deploy: add kustomize resources for deployment

WanzenBug:
Adds resources needed to deploy the operator itself. Most of it is taken as-is from the resources generated by operator-sdk.

The main changes:
* Deploy webhook configuration by default.
* Use cert-manager for webhook certificates by default.
* Add default_images.yaml as config map to operator.
* Use piraeus-datastore namespace.
* Remove LinstorSatellite sample: creation is reserved for the operator itself.

### DIFF
--- a/config/default/linstor_passphrase.yaml
+++ b/config/default/linstor_passphrase.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: linstor-controller-passphrase
+  namespace: system
+immutable: true
+data:
+  MASTER_PASSPHRASE: cGxlYXNlLWNoYW5nZS1tZQ== # please-change-me
--- a/config/default/linstor_passphrase.yaml
+++ b/config/default/linstor_passphrase.yaml
@@ -1,9 +1,0 @@
----
-apiVersion: v1
-kind: Secret
-metadata:
-  name: linstor-controller-passphrase
-  namespace: system
-immutable: true
-data:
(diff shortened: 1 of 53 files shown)


config/manifests/bases/piraeus-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,56 @@
...
+  keywords:
+  - storage
+  links:
+  - name: Piraeus Operator
+    url: https://piraeus-operator.domain
+  maturity: alpha
+  provider:
+    name: Piraeus Datastore
+    url: https://piraeus.io
+  version: 0.0.0
```
JoelColledge, line 56: Is this overridden somewhere?

WanzenBug: It gets overridden when running `make bundle`

config/default/linstor_passphrase.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: linstor-controller-passphrase
+  namespace: system
+immutable: true
+data:
+  MASTER_PASSPHRASE: cGxlYXNlLWNoYW5nZS1tZQ== # please-change-me
```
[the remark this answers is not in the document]
WanzenBug: Removed, because I don't like the solution here. Need to think about a better solution...

JoelColledge (approved):

PR #363, opened 2022-12-01
olm: ensure generated CSV passes scorecard tests

WanzenBug:
Ensures that the generated ClusterServiceVersion passes the basic "scorecard" tests. In particular, we need to ensure we have a basic description of the status field, as well as a description of all resources generated by a CR.

For the second task, we create a small program that collects that information directly from the resources embedded in the operator.


WanzenBug:
Made it a draft again, as I'm not quite happy with current display in OpenShift

hack/csvinflater/csvinflater.go
```diff
@@ -0,0 +1,165 @@
+package main
```
JoelColledge, line 1: Document what this is for. Especially because "CSV" has a different meaning in most of the computing world...

hack/csvinflater/csvinflater.go
```diff
@@ -0,0 +1,165 @@
...
+			Value: GetApiResources(satelliteRes.AllIds()...),
+		},
+	}
+
+	bytes, err := yaml.Marshal(patch)
+	if err != nil {
+		log.Fatalf("failed to marshal patch: %s", err)
+	}
+
+	fmt.Print(string(bytes))
```
JoelColledge, line 87: Also print that this is a generated file?

hack/csvinflater/csvinflater.go
WanzenBug: Good point.

JoelColledge (approved):

PR #364, opened 2022-12-02
cluster: add linstor passphrase to API

WanzenBug:
Add the option to reference a secret holding the LINSTOR master passphrase. The passphrase is then injected into the LINSTOR Controller pod via environment variables.

This is quite similar to how it worked in Operator v1. The main difference is that for v2, we do not plan to automatically create such a secret. This was always poorly thought out, especially as temporary removal of the deployment could lead to complete loss of the passphrase. By requiring the user to create one on their own, they hopefully keep better track of it (or use some external solution for storing it).

### DIFF
--- a/config/samples/linstorcluster.yaml
+++ b/config/samples/linstorcluster.yaml
@@ -6,16 +6,18 @@ spec: {}
 #  repository: registry.example.com
 #  nodeSelector:
 #    piraeus.io/satellite: "true"
+#  linstorPassphraseSecret: my-linstor-passphrase
+#  properties:
+#    - name: TcpPortAutoRange
+#      value: "8000-8999"
 #  patches:
 #    - target:
-#        kind: Pod
-#        name: satellite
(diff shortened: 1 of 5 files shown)


config/samples/linstorcluster.yaml
```diff
@@ -6,16 +6,18 @@ spec: {}
 #  repository: registry.example.com
 #  nodeSelector:
 #    piraeus.io/satellite: "true"
+#  linstorPassphraseSecret: my-linstor-passphrase
+#  properties:
+#    - name: TcpPortAutoRange
+#      value: "8000"
```
JoelColledge, line 12: Only just noticed this detail which is also in the README: Does this actually work? Shouldn't the value be "8000-9999" or something of that form?

WanzenBug: Yes, you are right.

JoelColledge (approved):

PR #365, opened 2022-12-05
controller: enable leader election

WanzenBug:
Leader election is used to ensure only one LINSTOR Controller is running
at any single point in time. We only use the basic form of leader election
without manually configuring the Kubernetes service: We rely on a single
replica on the Deployment, with shortened tolerations for disconnected
nodes.

This means we don't have to rely on manually updating the Service,
while still having reasonably fast fail-over times.

### DIFF
--- a/pkg/resources/cluster/controller/controller-deployment.yaml
+++ b/pkg/resources/cluster/controller/controller-deployment.yaml
@@ -26,6 +26,20 @@ spec:
           env:
             - name: JAVA_OPTS
               value: '-Djdk.tls.acknowledgeCloseNotify=true'
+            - name: K8S_AWAIT_ELECTION_ENABLED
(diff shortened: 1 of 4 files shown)


JoelColledge (dismissed):
Code looks fine. Commit message corrections:

```
Leader election is used to ensure only one LINSTOR Controller is running
at any single point in time. We only use the basic form of leader election
without manually configuring the Kubernetes service: We rely on a single
replica on the Deployment, with shortened tolerations for disconnected
nodes.

This means we don't have to rely on manually updating the Service,
while still having reasonably fast fail-over times.
```

WanzenBug:
> Commit message corrections

Thanks, updated.

JoelColledge (approved):

PR #367, opened 2022-12-12
v2: LINSTOR internal TLS

WanzenBug:
Add a new field to all CRDs, allowing users to configure TLS certificates.

There are two ways to configure TLS certificates:
* Manual provisioning. In this case the operator will assume that all secrets already exist. The user has to ensure that the Clusters certificate is trusted by the satellites and vice versa.
* `cert-manager` based provisioning. In this case the operator will create [`Certificate`](https://cert-manager.io/docs/usage/certificate/) resources using an `Issuer` provided by the user.

The user can omit explicitly naming any secrets, in which case default values of `linstor-controller-internal-tls` and `<node-name>-tls` will be used.

In addition to the above, also fix a small issue related to showing errors in the status field


api/v1/linstorsatelliteconfiguration_types.go
```diff
@@ -47,6 +47,12 @@ type LinstorSatelliteConfigurationSpec struct {
 	// +patchMergeKey=name
 	// +patchStrategy=merge
 	Properties []LinstorNodeProperty `json:"properties,omitempty"`
+
+	// InternalTLS configures secure communication for the LINSTOR Satellite.
+	//
+	// If set, the control traffic between LINSTOR Controller and Satellite will be encrypted using mTLS.
+	// +kubebuilder:validation:Optional
+	InternalTLS *TLSConfig `json:"internalTLS,omitempty"`
```
JoelColledge, line 55: Does this one not need `+nullable`?

api/v1/linstorsatellite_types.go
```diff
@@ -46,6 +46,15 @@ type LinstorSatelliteSpec struct {
 	// +patchMergeKey=name
 	// +patchStrategy=merge
 	Properties []LinstorNodeProperty `json:"properties,omitempty"`
+
+	// InternalTLS configures secure communication for the LINSTOR Satellite.
+	//
+	// If set, the control traffic between LINSTOR Controller and Satellite will be encrypted using mTLS.
+	// The Controller will use the client key from `LinstorCluster.spec.internalTLS` when connecting.
+	// +kubebuilder:validation:Optional
+	// + Without "nullable" the k8s API does not accept patches with 'internalTLS: {}', which seems to be a bug.
```
JoelColledge, line 55: Is the `+` necessary?

```suggestion
	// Without "nullable" the k8s API does not accept patches with 'internalTLS: {}', which seems to be a bug.
```

README.md
```diff
@@ -75,6 +75,20 @@ The following customizations are available in the `LinstorCluster` resource:
   spec:
     linstorPassphraseSecret: my-linstor-passphrase
   ```
+* A reference to a TLS secret, containing the secret key and certificate use in mutual authentication of  the satellites.
```
JoelColledge, line 78: We should document the implicit naming of secrets and show how to configure TLS without `cert-manager`.

controllers/linstorsatellite_controller.go
```diff
@@ -530,3 +538,122 @@ func (r *LinstorSatelliteReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}, applycorev1.Pod("satellite", "").
+		WithSpec(applycorev1.PodSpec().
+			WithAffinity(applycorev1.Affinity().
+				WithNodeAffinity(applycorev1.NodeAffinity().
+					WithRequiredDuringSchedulingIgnoredDuringExecution(applycorev1.NodeSelector().
+						WithNodeSelectorTerms(applycorev1.NodeSelectorTerm().
+							WithMatchFields(applycorev1.NodeSelectorRequirement().
+								WithKey("metadata.name").
+								WithOperator(corev1.NodeSelectorOpIn).
+								WithValues(nodeName),
```
JoelColledge, line 554: Is it deliberate that this is now `lsatellite.Name` rather than `node.Name`? Didn't we have the case that the underlying node may change name but the satellite needs to stay on that node with the old name. So we need to match the new node name.

api/v1/linstorsatelliteconfiguration_types.go
WanzenBug: Not in my testing. `nullable` seems to be only needed because we use "server-side apply" and at some point the API server converts a `internalTLS: {}` into `internalTLS: null`, which then triggers validation. I only added it to the satellite because that is the one point where the operator can trigger this behaviour.

Thinking about it some more, we could also add it here and in the LinstorCluster CRD, should any of our users also use server-side apply for their own configuration :thinking:

api/v1/linstorsatellite_types.go
WanzenBug: Yes, this hides the content from the auto-generated docs of the CRDs, where I think this comment isn't really useful.

controllers/linstorsatellite_controller.go
WanzenBug: I don't think a node in kubernetes can change name. Specifically, in Operator v2, satellite-name == k8s-node-name, always. We can use those interchangeably.

README.md
WanzenBug: Good point. I'm wondering if this would be a good time to think about proper splitting for our documentation (see also https://diataxis.fr/)

The README is already quite long, with lots of information that's probably not needed, while actual reference documentation is not available.

JoelColledge: Sounds good. Do you have a plan already for implementing that?

We can merge this PR and then work on the docs.

JoelColledge (approved):

README.md
WanzenBug: Having it written down will surely help, but now I think we need ChatGPT to generate all that text :sweat_smile: https://github.com/piraeusdatastore/piraeus-operator/issues/368

PR #373, opened 2022-12-19
csi: restart node pods if topology labels changed

WanzenBug:
Reimplementation of 1b1e30d2e74e.
    
The satellite operator applies k8s node labels as aux properties. These are
then reported by the csi nodes as supported topology keys. This report
only happens once on plugin startup. Later changes are not reflected, leaving
the volume provisioning steps with old or outdated topology keys.
    
The only way to force a re-sync of those keys is to delete pod. This change
compares the LINSTOR Node auxilliary properties with the keys reported by
the csi plugin registration. If the set of labels has diverged, the
registration and pod is deleted. After the daemonset restarts the pod,
kubelet will trigger a new registration, making the updated keys available.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -61,6 +61,14 @@ rules:
   verbs:
   - patch
 - apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - delete
+  - list
+  - watch
+- apiGroups:
   - apiextensions.k8s.io
   resources:
   - customresourcedefinitions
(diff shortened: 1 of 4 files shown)


WanzenBug:
Note: I thought about adding a test case, but couldn't come up with a way to test without having to mock the linstor client.

JoelColledge (dismissed):
Looks good. I think it could be improved with a brief comment explaining why we do this. It is not obvious and the reader shouldn't have to delve into git history to understand the purpose of the code.

JoelColledge (approved):

PR #377, opened 2022-12-27
satellite: add file backed storage pools

WanzenBug:
LINSTOR can use a normal file system to create a storage pool. This commit adds support for this in the operator:

* A new option in the storage pool section
* Automatic mount of the host directory
* Default values if only a storage pool name was specified
* Reference documentation

Note: currently does not survive a pod restart. This can be fixed in the future by having the LINSTOR Satellite use an enhanced version of `losetup`


docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -78,6 +78,10 @@ the matching key. Available types are:
 * `lvmThin`: Configures a [LVM Thin Pool](https://man7.org/linux/man-pages/man7/lvmthin.7.html) as storage pool.
   Defaults to using the storage pool name as name for the thin pool volume and the storage pool name prefixed by
   `linstor_` as the VG name. Can be overridden by setting `thinPool` and `volumeGroup`.
+* `file`: Configures a file system based storage pool. Configures a host directory as location for the volume files.
```
JoelColledge, line 81: We need to make it extra clear that `file:` contains the path to a directory, if set. That is a bit counter-intuitive. One would expect something named "file" to be a filename or a path to a file.

We could even change the key to `filePool` or something to clarify this.

Perhaps include an example with a path too?

WanzenBug: It would currently look like this:

```
file:
  directory: /var/lib/linstor-pools/foo
```

But I agree that is confusing. I somehow don't like `filePool`, because then you have the repeated pool `storagePools: [ filePool: ... ]`, but that is a minor issue. Would `fileSystem` be an alternative?

JoelColledge: > It would currently look like this:

That is at least a bit clearer than I had thought.

`fileSystem` could work. `fileBacked`? Is there any precedent in the linstor docs? I can't see any with a quick search.

WanzenBug: > Is there any precedent in the linstor docs

Doesn't look like it:

> Create a file storage pool

and

> File / FileThin: Mostly used for demonstration / experiments. LINSTOR will basically reserve a file in a given directory and will configure a [loop device](https://man7.org/linux/man-pages/man4/loop.4.html) on top of that file.

I don't think there is anything to take from that.

---

`fileBacked` sounds good.

Would it then be `fileBackedThin`, `thinFileBacked`, `sparseFileBacked`, `fileThinBacked`? I don't really like any of these. Perhaps it's better to have it as an option on the `fileBacked` struct? I just mirrored the way it is configured for LVM, because the options there are actually different.

JoelColledge: My preference is actually `filePool` and `fileThinPool`. That avoids introducing a new term in this context. Shame about the aesthetics :man_shrugging:

docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -78,6 +78,10 @@ the matching key. Available types are:
 * `lvmThin`: Configures a [LVM Thin Pool](https://man7.org/linux/man-pages/man7/lvmthin.7.html) as storage pool.
   Defaults to using the storage pool name as name for the thin pool volume and the storage pool name prefixed by
   `linstor_` as the VG name. Can be overridden by setting `thinPool` and `volumeGroup`.
+* `filePool`: Configures a file system based storage pool. Configures a host directory as location for the volume files.
+  Defaults to using the `/var/lib/linstor-pools/<storage pool name>` directory.
+* `fileThinPool`: Configures a file system based storage pool. Behaves the same as `fileSystem`, except the files will
```
JoelColledge, line 83:
```suggestion
* `fileThinPool`: Configures a file system based storage pool. Behaves the same as `filePool`, except the files will
```

docs/reference/linstorsatelliteconfiguration.md
WanzenBug: > Maybe we should do that

Done

JoelColledge (dismissed):
Looks good to me. I'll leave it open until you are happy with the end-to-end tests.

WanzenBug:
One issue I noticed during testing: If a user uses something like `MyPool` as the name for the file pool, the operator will try to create the following Pod:

```yaml
spec:
  containers:
  - name: linstor-satellite
    volumeMounts:
    - name: MyPool
      mountPath: /var/lib/linstor-pools/MyPool
  volumes:
  - name: MyPool
    hostPath:
      path: /var/lib/linstor-pools/MyPool
 ...
```

The issue is that `MyPool` is not a valid volume identifier in k8s. For some reason it must be a [RFC1123](https://www.rfc-editor.org/rfc/rfc1123) "hostname", i.e. only `a-z + "-"` are allowed. We could change it so that `MyPool` get's converted to `mypool` or maybe even `my-pool`. Or we could just name the volumes `volume1`, `volume2` etc...

JoelColledge:
> For some reason it must be a [RFC1123](https://www.rfc-editor.org/rfc/rfc1123) "hostname"

:facepalm:

> We could change it so that MyPool get's converted to mypool or maybe even my-pool.

Sounds fine. Just have to watch out that the mapping is injective.

> Or we could just name the volumes volume1, volume2 etc...

Sounds good enough to me. Maybe `linstor-pool-0` or `file-pool-0`.

JoelColledge (approved):

PR #380, opened 2023-01-03
controller: requeue reconcile in intervals

WanzenBug:
Since we don't get notifications from LINSTOR Controller for all relevant events, we have to ensure we requeue reconcile attempts from time to time.

One particular occurance where this is relevant are starting LINSTOR Satellites: since the Pod can be ready long before LINSTOR reconnects, we might miss the window to update the status once the satellite is actually online.

### DIFF
--- a/pkg/conditions/conditions.go
+++ b/pkg/conditions/conditions.go
@@ -105,3 +105,13 @@ func (c Conditions) ToConditions(generation int64) []metav1.Condition {
 
 	return result
 }
+
+func (c Conditions) AllHaveStatus(status metav1.ConditionStatus) bool {
+	for _, cond := range c {
+		if cond.Status != status {
+			return false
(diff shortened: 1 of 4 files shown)


pkg/conditions/conditions.go
```diff
@@ -105,3 +105,13 @@ func (c Conditions) ToConditions(generation int64) []metav1.Condition {
 
 	return result
 }
+
+func (c Conditions) AllOfStatus(status metav1.ConditionStatus) bool {
```
JoelColledge, line 109: The name is a bit confusing. "All of..." is normally used in the sense of "all of the members", so you'd expect it to return a collection. `AllHaveStatus`? `AllStatusEqual`?

JoelColledge (approved):

PR #381, opened 2023-01-04
patches: convert to yaml-based patches

WanzenBug:
Having patches to k8s resources directly in the controller code made the code quite hard to follow.

To replace those patches, use predefined patches, already in the kustomize format. Since the patches only need setting of single values, we create a basic "yaml-safe" template system, where we can replace a specific part of the patch with minimal fuzz.

This commit also converts almost all of the patches to this new system. Only the LinstorSatellite patch created by the LinstorCluster controller is left as the old set of manual JSONPatch operations, as it does not fit nicely in the "replace single parts of patches" framework.

### DIFF
--- a/controllers/linstorsatellite_controller.go
+++ b/controllers/linstorsatellite_controller.go
@@ -17,17 +17,14 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"net"
 	"strings"
 	"time"
 
-	"github.com/BurntSushi/toml"
 	linstor "github.com/LINBIT/golinstor"
 	lclient "github.com/LINBIT/golinstor/client"
-	"github.com/LINBIT/golinstor/linstortoml"
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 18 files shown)


WanzenBug:
@JoelColledge oops, I wrote a templated YAML patches :sweat_smile: 

Overall I'm quite happy with this system. I guess it's a bit inefficient converting from YAML, crawling the maps and converting back. I think it's worth it just to have the patches in the same format as the resources.

controllers/linstorsatellite_controller.go
```diff
@@ -204,40 +200,22 @@ func (r *LinstorSatelliteReconciler) kustomizeNodeResources(lsatellite *piraeusi
...
-					Path:  "/metadata/name",
-					Value: secretName,
-				},
-				{
-					Op:    utils.Replace,
-					Path:  "/spec/secretName",
-					Value: secretName,
-				},
-			})
+			p, err := SatelliteLinstorInternalTLSCertManagerPatch(secretName, lsatellite.Spec.InternalTLS.CertManager)
```
JoelColledge, line 213: I don't quite see the benefit of converting these `JsonPatch` ones. They use a different format to the yaml resources anyway.

JoelColledge (commented):
Neat!

I haven't checked the individual patches yet.

controllers/linstorsatellite_controller.go
WanzenBug: You are right. And right now I don't know why they are `JsonPatch` in the first place.

WanzenBug: Replaced those patches with strategic merge ones. I only  used it because I thought I needed to change the names, but that is either:
* better done using `replacements`
* just not needed

JoelColledge (approved):

PR #383, opened 2023-01-05
secure LINSTOR API using TLS

WanzenBug:
Add support for a secure LINSTOR API using mTLS. The design is similar to the existing "internalTLS" setup used for securing the Controller <-> Satellite communication.

The user can set a single field in the LinstorCluster spec named "apiTLS" to enable the protected API. There are a collection of options on the object, like the secret names and optional support for secret provisioning via cert-manager.

The reason for having multiple secrets, and not just one server + one client is to future-proof the design should LINSTOR eventually gain proper authentication support: not every component needs access to the full API. For example, the CSI Node only needs read-only access to just the volume API. In the future we might want to distinguish between these API clients.

For the operator itself, we also need access to the API, both from the LinstorCluster controller as well as the LinstorSatellite controller. Since the LinstorSatellite controller does not have direct access to the LinstorCluster, we add the secret name to the ClusterReference. This makes it easy to read the client certificate from the secret.


api/v1/linstorcluster_types.go
```diff
@@ -65,6 +66,74 @@ type LinstorClusterSpec struct {
...
+	// Defaults to "linstor-api-tls".
+	//+kubebuilder:validation:Optional
+	ApiSecretName string `json:"apiSecretName,omitempty"`
+
+	// ClientSecretName references a secret holding the TLS key and certificate used to by the operator to configure
+	// the cluster. Defaults to "linstor-client-tls".
+	//+kubebuilder:validation:Optional
+	ClientSecretName string `json:"clientSecretName,omitempty"`
+
+	// CsiControllerSecretName references a secret holding the TLS key and certificate used to by the CSI Controller
```
JoelColledge, line 90:
```suggestion
	// CsiControllerSecretName references a secret holding the TLS key and certificate used by the CSI Controller
```

api/v1/linstorcluster_types.go
```diff
@@ -65,6 +66,74 @@ type LinstorClusterSpec struct {
...
+	// to provision volumes. Defaults to "linstor-csi-controller-tls".
+	//+kubebuilder:validation:Optional
+	CsiControllerSecretName string `json:"csiControllerSecretName,omitempty"`
+
+	// CsiNodeSecretName references a secret holding the TLS key and certificate used to by the CSI Nodes to query
```
JoelColledge, line 95:
```suggestion
	// CsiNodeSecretName references a secret holding the TLS key and certificate used by the CSI Nodes to query
```

api/v1/linstorcluster_types.go
```diff
@@ -65,6 +66,74 @@ type LinstorClusterSpec struct {
...
+	// the volume state. Defaults to "linstor-csi-node-tls".
+	//+kubebuilder:validation:Optional
+	CsiNodeSecretName string `json:"csiNodeSecretName,omitempty"`
+
+	// CertManager references a cert-manager Issuer or ClusterIssuer.
+	// If set, a Certificate resources will be created, provisioning the secrets referenced in *SecretName using the
```
JoelColledge, line 101:
```suggestion
	// If set, a Certificate resource will be created, provisioning the secrets referenced in *SecretName using the
```

JoelColledge (commented):
Looks fine to me overall.

JoelColledge (approved):

PR #384, opened 2023-01-09
docs: add how-to for internal TLS

WanzenBug:
Add a how-to for setting up internal TLS, both with cert-manager and manually, using openssl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,12 @@ linstorsatelliteconfigurations.piraeus.io/all-satellites created
 
 ## Configuration
 
-We are currently working on improving our documentation. Additional tutorials, how-to guides and explanation will be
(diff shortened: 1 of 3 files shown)


WanzenBug:
Tried my hand at a first "how-to". Still a draft, because I haven't actually verified that the instructions are working, especially the manual TLS steps.

JoelColledge (commented):
Looks great! Reads as I would expect a how-to to.

WanzenBug:
Perhaps surprisingly, the instructions worked almost flawlessly. Only needed to add some namespace and other metadata.

JoelColledge (approved):

JoelColledge:
We should refer to this from somewhere so that users can find it.

JoelColledge (approved):

PR #405, opened 2023-02-13
tutorial: snapshots

WanzenBug:
Add a simple tutorial showing snapshot + restore

Also uses `1Gi` as volume size in doc examples consistently.


JoelColledge (approved):
Great!

PR #407, opened 2023-02-17
v2: reintroduce a helm chart

WanzenBug:
Reintroduce a Helm Chart, for users that really need it. We don't want to put too much effot into it, so most of it was created by running https://github.com/arttor/helmify once.

We have a separate README for the Chart, but otherwise avoid mentioning Helm in our main guides: new users should be using kustomize.

### DIFF
--- a/charts/piraeus/README.md
+++ b/charts/piraeus/README.md
@@ -1,0 +1,35 @@
+# Piraeus Operator
+
+Deploys the [Piraeus Operator](https://github.com/piraeusdatastore/piraeus-operator) which deploys and manages a simple
+and resilient storage solution for Kubernetes.
+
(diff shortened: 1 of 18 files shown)


charts/piraeus/README.md
```diff
@@ -0,0 +1,35 @@
+# Piraeus Operator
+
+Deploys the [Piraeus Operator](https://github.com/piraeusdatastore/piraeus-operator) which deploys and manages a simple
+and resilient storage solution for Kubernetes.
+
+The main deployment method for Piraeus Operator switched to [`kustomize`](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/docs/tutorial)
```
JoelColledge, line 6: Use relative links to avoid hardcoding `v2` path which may change?

JoelColledge (approved):

PR #413, opened 2023-02-23
v2.0.0 release

WanzenBug:
Fixes some issues in the release process compared to v2.0.0-rc.1, otherwise pretty much unchanged.


JoelColledge:
We only just merged rc.1 so I'm a bit surprised to get the final release PR already. What's the plan?

WanzenBug:
Plan is to get this out the door. I only did the rc.1to test that the release infrastructure actually works as expected.

People didn't seem to test our RCs in the past, so I don't see too much benefit having a longer RC phase.

JoelColledge:
> People didn't seem to test our RCs in the past, so I don't see too much benefit having a longer RC phase.

In this case it certainly makes sense. Users who need stability will stick on `v1` for now anyway.

JoelColledge (approved):

PR #419, opened 2023-03-06
satellite: add drbd-shutdown-guard init-container

WanzenBug:
The purpose of this container is to install a systemd service unit that runs during normal shutdown. When the unit is stopped, it will force all resources into secondary role. This will cause any suspended resource to get "unstuck" for the purpose of unmounting.

The service is congigured to stop after kubelet, so we can assume that all still mounted DRBD resources will remain so until systemd does it's normal clean up. It also means that the network for DRBD is likely already torn down, so any pending writes would be lost in any case.

### DIFF
--- a/pkg/resources/satellite/patches/precompiled-module.yaml
+++ b/pkg/resources/satellite/patches/precompiled-module.yaml
@@ -10,16 +10,11 @@
       name: satellite
     spec:
       volumes:
-      - name: lib-modules
-        $patch: delete
       - name: usr-src
         $patch: delete
       initContainers:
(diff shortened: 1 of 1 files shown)


JoelColledge (commented):
Looks OK to me. It is a daring concept, but I understand the necessity for it. Has anyone reviewed `drbd-shutdown-guard` itself yet?

WanzenBug:
> Has anyone reviewed drbd-shutdown-guard itself yet?

Yes, Roland did a review of it.

JoelColledge (approved):

PR #421, opened 2023-03-08
Release 2.0.1

WanzenBug:
Release some accumulated bug fixes since 2.0.0.

Technically, because of the addition of `drbd-shutdown-guard`, this should be 2.1.0. But that somehow feels wrong. I can convince myself this is a "fix" for stuck shutdowns...


JoelColledge (approved):

JoelColledge:
> Technically, because of the addition of drbd-shutdown-guard, this should be 2.1.0.

I would have gone for 2.1.0. I can imagine that `drbd-shutdown-guard` could cause problems in some obscure environment that we haven't considered. If someone wanted a very stable system and only wanted to accept bug fixes then they wouldn't want to add that.

However, I don't think anyone is in this situation in practice. Especially because v2 is so new.

PR #444, opened 2023-03-28
Add Debian support

clrxbl:
This adds drbd-driver/module-loader support for Debian 10 & Debian 11.
/usr/lib has to be mounted in the loader container, as Debian symlinks a lot of kernel header stuff e.g. Makefile scripts back into /usr/lib.

This depends on https://github.com/piraeusdatastore/piraeus/pull/138

### DIFF
--- a/charts/piraeus/templates/config.yaml
+++ b/charts/piraeus/templates/config.yaml
@@ -54,3 +54,7 @@ data:
             image: drbd9-focal
           - osImage: Ubuntu 22\.04
             image: drbd9-jammy
+          - osImage: Debian GNU/Linux 11 (bullseye)
+            image: drbd9-bullseye
+          - osImage: Debian GNU/Linux 10 (buster)
+            image: drbd9-buster
--- a/charts/piraeus/templates/config.yaml
+++ b/charts/piraeus/templates/config.yaml
@@ -54,7 +54,7 @@ data:
             image: drbd9-focal
           - osImage: Ubuntu 22\.04
             image: drbd9-jammy
-          - osImage: Debian GNU/Linux 11 (bullseye)
+          - osImage: Debian GNU/Linux 11
             image: drbd9-bullseye
-          - osImage: Debian GNU/Linux 10 (buster)
+          - osImage: Debian GNU/Linux 10
             image: drbd9-buster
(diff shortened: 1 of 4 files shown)


pkg/resources/satellite/pod/node-pod.yaml
```diff
@@ -29,6 +29,9 @@ spec:
         - mountPath: /usr/src
           name: usr-src
           readOnly: true
+        - mountPath: /usr/lib
+          name: usr-lib
+          readOnly: true
```
WanzenBug, line 34: I wonder why this is needed for debian, but not for any of the other distributions. They should all have `/lib -> /usr/lib` links :thinking:

charts/piraeus/templates/config.yaml
```diff
@@ -54,3 +54,7 @@ data:
             image: drbd9-focal
           - osImage: Ubuntu 22\.04
             image: drbd9-jammy
+          - osImage: Debian GNU/Linux 11 (bullseye)
+            image: drbd9-bullseye
+          - osImage: Debian GNU/Linux 10 (buster)
```
WanzenBug, line 59: I'd use a more general pattern, since these are regular expressions anyways. Perhaps just `bullseye` and `buster` are enough?

charts/piraeus/templates/config.yaml
```diff
@@ -54,3 +54,7 @@ data:
...
+            image: drbd9-buster
```
WanzenBug, line 60: Also: missing newline at the end of file. Consider installing https://pre-commit.com/, `pre-commit run` should then fix the file automatically.

WanzenBug (changes requested):

charts/piraeus/templates/config.yaml
clrxbl: I'd be more inclined to use Debian GNU/Linux 11 and strip away the codename in brackets. Not fully sure how Kustomize handles it.

pkg/resources/satellite/pod/node-pod.yaml
WanzenBug: :bulb: That's the missing puzzle piece. So we need additional tools because debian splits some things off into `/usr/lib/linux-kbuild-X`. On the other distros, `/lib/modules` and `/usr/src` is enough.

charts/piraeus/templates/config.yaml
WanzenBug: > strip away the codename

Also fine by me. It should "just work".

WanzenBug (approved):
LGTM for now. Perhaps in the future we improve this even more, with every distro only having to bind-mount what's actually required.

PR #451, opened 2023-04-05
Support external LINSTOR Controllers

### LINKED ISSUE #447: Piraeus operators connected to external linstor storage clusters
I have an external linstor storage cluster
Can I use piraeus-operator to connect to it
piraeus-operator is only responsible for running satellite nodes

WanzenBug:
Allow users to specify the URL of an external LINSTOR Controller. When specified, no LINSTOR Controller is deployed. Instead the operator will reconcile the Kubernetes resources so that they use the external controller instead.

Fixes #447 for v2


pkg/linstorhelper/client.go
```diff
@@ -58,31 +78,27 @@ func NewClientForCluster(ctx context.Context, cl client.Client, namespace, clust
...
-			return "https", p.Port
-		}
-
-		if p.Name == "api" {
-			port = p.Port
+		switch p.Name {
+		case "secure-api":
+			return "https", p.Port, true
+		case "api":
+			return "http", p.Port, true
```
JoelColledge, line 97: If `svc.Spec.Ports` contains an `"api"` port followed by an `"secure-api"` port, this now returns the former instead of the latter. I'm guess this behavior change is deliberate?

docs/how-to/external-controller.md
```diff
@@ -0,0 +1,87 @@
...
+After applying the `LinstorCluster`, Piraeus will make the following changes to the default deployment:
+
+* The `linstor-controller` Deployment is removed
+* The `linstor-csi-controller` and `linstor-csi-node` resources will reference the external controller.
+
+## Configuring Host Networking for LINSTOR Satellites
+
+Next, configure your Satellites to use [host networking](./drbd-host-networking.md). Unless your Pod network is
+reachable from outside the Kubernetes Cluster, the existing LINSTOR Cluster won't be able to communicate with the
+Satellites in the Kubernetes cluster.
```
JoelColledge, line 33: Slightly simpler logical flow IMHO:

```suggestion
Normally the Pod network is not reachable from outside the Kubernetes Cluster.
In this case the existing LINSTOR Controller won't be able to communicate with the Satellites in the Kubernetes cluster.
So you should configure your Satellites to use [host networking](./drbd-host-networking.md).
```

JoelColledge (commented):
Looks great to me.

docs/how-to/external-controller.md
WanzenBug: Thanks, applied!

pkg/linstorhelper/client.go
WanzenBug: Not really, was just easier to implement that way. Changed it back to prefer `"secure-api"` if both are available, while still reporting missing services.

JoelColledge (approved):

PR #452, opened 2023-04-05
Add info about kernel module config to talos docs

mcflis:
Adds [`KernelConfig`](https://www.talos.dev/v1.3/reference/configuration/#kernelconfig) info to the Talos docs.

Without it, `talosctl -n <NODE_IP> read /proc/modules` won't return anything and the linstor controller will report DRBD to be unsupported:
```
Unsupported resource layers:
 <node-name>:
  DRBD: DRBD version has to be >= 9. Current DRBD version: 0.0.0
```

> source: https://github.com/siderolabs/talos/pull/6426/files#diff-1a8257ae3f3e1367538a005610365d7fe5895c5833f1b802d4a88b9d3433c597R126-R129

### DIFF
--- a/docs/how-to/talos.md
+++ b/docs/how-to/talos.md
@@ -24,6 +24,8 @@ machine:
   kernel:
     modules:
       - name: drbd
+        parameters:
+          - usermode_helper=disabled
       - name: drbd_transport_tcp
 ```
 **NOTE**: Replace `v1.3.6` with the Talos version running.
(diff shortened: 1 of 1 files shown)


docs/how-to/talos.md
```diff
@@ -20,14 +20,19 @@ This can be achieved by updating the machine config:
 machine:
   install:
     extensions:
-      - image: ghcr.io/siderolabs/drbd:9.2.0-v1.3.5
+      - image: ghcr.io/siderolabs/drbd:9.2.0-v1.3.6
+  kernel:
+    modules:
+      - name: drbd
```
WanzenBug, line 26: Can one add some parameters here? If so, the kernel module should be loaded with `usermode_helper=disabled`

WanzenBug (changes requested):
Thanks for jumping in and fixing the documentation!

docs/how-to/talos.md
mcflis: done: 431f26bdd43599913c57b8ec1379a50a85c8619e

WanzenBug (approved):

WanzenBug:
Thanks again for contributing!

PR #455, opened 2023-04-11
how-to: run on MicroK8s

WanzenBug:
MicroK8s needs slightly modified mount paths for the CSI node containers.

During testing it also revealed an issue in the way we pass the csi socket to the linstor-csi plugin: instead of using the explicit mount location it used a path for the csi socket that only worked by accident. To fix it, we now use the explicit socket mount location.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ability to skip deploying the LINSTOR Controller by setting `LinstorCluster.spec.externalController`.
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


JoelColledge (approved):

JoelColledge:
Do we not want `--csi-endpoint` to be configurable?

WanzenBug:
> Do we not want --csi-endpoint to be configurable?

No, it's just an implementation detail in this case. It would only make sense to make `--kubelet-registration-path=/var/lib/kubelet/plugins/linstor.csi.linbit.com/csi.sock` configurable, as that is the option that sets the path that `/csi/csi.sock` is exposed as on the host. But no need change it here.

Issue #424, opened 2023-03-10
StoragePools not recovering after node loss - Self Heal Option?


inviscid:
I'm not sure if I've have everything configured correctly since my StoragePools do not recover when nodes are lost.  Since a node has no guarantee to have the same name after a loss (GKE node creation), the StoragePool continues to look for the original node name and goes into a state where other nodes are now available (the replacements) but continues to think the original nodes will come back online.

Is there a way to configure Piraeus to self-heal the StoragePools using whatever available nodes meet the NodeSelector criteria?  If nodes are offline for some period of time can they be automatically removed from the StoragePool since they are likely gone forever.

inviscid:
After a little more investigation I see that even for the nodes that come back up with the same name, the `linstor-csi-node` daemonset never enters a `Ready` state because the `linstore-wait-node-online` container keeps thinking the node is not online.  

```
time="2023-03-10T15:04:03Z" level=info msg="not ready" error="satellite gke-gp-test2-greenplum-165811ea-3g2x is not ONLINE: OFFLINE" version=refs/tags/v0.2.1
time="2023-03-10T15:04:13Z" level=info msg="not ready" error="satellite gke-gp-test2-greenplum-165811ea-3g2x is not ONLINE: OFFLINE" version=refs/tags/v0.2.1
```

Without the `linstor-csi-node` daemonset running, I suspect this is why Piraeus can't recover the StoragePool properly after node losses.  I'm not sure why the `linstore-wait-node-online` container thinks the node is offline when it appears to be online as far as I can tell.

inviscid:
I discovered the problem is when the `LinstorSatellite` pod fails, there is no automated recovery.  The pod remains in a failed state until manual intervention.  Since the `LinstorSatellite`  `linstor-satellte` container isn't online, the `linstor-csi-node` remains in a pending state.

Should the `LinstorSatellite` try to self-heal so manual intervention is not required?

inviscid:
Besides the self healing option above when pods fail, I also discovered that pods are not being rescheduled to available nodes with a replica of the data.  Is there an additional scheduler that must be specified to allow pod scheduling on a node with one of its replicas?

WanzenBug:
Strange, the Operator should actually try to recreate the Satellite Pods if they enter a failed state. Perhaps there is a bug in the logic there. What's the state of the Pods after rotating nodes?

[...]

PR #458, opened 2023-04-17
operator: run reconiliation on changed image config

WanzenBug:
The image configuration can be changed without the Operator restarting. But the Operator would never reload the config, so the changes would only take effect after restarting the Pod.

This commit adds support for reading the image configuration directly from the kubernetes API, also triggering a reload automatically when a changed config is detected.


WanzenBug:
This was inspired by #424: Changing the config map should trigger a change in deployment.

JoelColledge (approved):
All looks good to me.

PR #465, opened 2023-04-26
controller: add automatic backup for database migration

WanzenBug:
Add a new initContainer to the LINSTOR Controller deployment. The command will ensure that the k8s DB used by LINSTOR is up-to-date before the real LINSTOR Controller starts.

The "runMigration" command used will also create a backup of existing DB resources before running any migrations.

This needs https://github.com/piraeusdatastore/piraeus/pull/143


pkg/resources/cluster/controller/controller-deployment.yaml
```diff
@@ -18,6 +18,37 @@ spec:
...
+            - name: K8S_AWAIT_ELECTION_IDENTITY
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+          securityContext:
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
+          volumeMounts:
```
JoelColledge, line 46: The container references `/var/log/linstor-controller`. Does this need to be mounted?

WanzenBug: No, the `linstor-config` tool only logs to stdout/stderr.

JoelColledge (approved):

PR #468, opened 2023-05-05
Implement rate limiter for LINSTOR API

WanzenBug:
LINSTOR may get overloaded on large clusters once the operator starts up. To combat this, we implement a global rate limit for the LINSTOR API, defaulting to 100 qps.

While at it, also set a user agent identifying the operator.


JoelColledge (approved):
Neat.

Issue #427, opened 2023-03-13
drbd-module-loader container


lysyy:
Debian 11
Need a git checkout to regenerate drbd/.drbd_git_revision
make[1]: Entering directory '/tmp/pkg/drbd-9.2.2/drbd'

    Calling toplevel makefile of kernel source tree, which I believe is in
    KDIR=/lib/modules/5.10.0-19-amd64/build

make -C /lib/modules/5.10.0-19-amd64/build   M=/tmp/pkg/drbd-9.2.2/drbd  modules
arch/x86/Makefile:147: CONFIG_X86_X32 enabled but no binutils support
make[3]: gcc-10: No such file or directory
/usr/src/linux-headers-5.10.0-19-common/Makefile:983: /usr/src/linux-headers-5.10.0-19-common/scripts/Makefile.extrawarn: No such file or directory
make[3]: *** No rule to make target '/usr/src/linux-headers-5.10.0-19-common/scripts/Makefile.extrawarn'.  Stop.
make[2]: *** [/usr/src/linux-headers-5.10.0-19-common/Makefile:185: __sub-make] Error 2
make[1]: *** [Makefile:134: kbuild] Error 2
make[1]: Leaving directory '/tmp/pkg/drbd-9.2.2/drbd'
make: *** [Makefile:126: module] Error 2

Could not find the expexted *.ko, see stderr for more details

WanzenBug:
This is because we don't have a matching image here: https://github.com/piraeusdatastore/piraeus/tree/master/dockerfiles/drbd-driver-loader

And we don't detect any debian OS versions here: https://github.com/piraeusdatastore/piraeus-operator/blob/v2/config/manager/default_images.yaml#L34

To give some insight: LINBIT does not officially support DRBD 9 on Debian, so it will remain untested.

clrxbl:
> This is because we don't have a matching image here: [piraeusdatastore/piraeus@`master`/dockerfiles/drbd-driver-loader](https://github.com/piraeusdatastore/piraeus/tree/master/dockerfiles/drbd-driver-loader?rgh-link-date=2023-03-13T12%3A48%3A27Z)
> 
> And we don't detect any debian OS versions here: [`v2`/config/manager/default_images.yaml#L34](https://github.com/piraeusdatastore/piraeus-operator/blob/v2/config/manager/default_images.yaml?rgh-link-date=2023-03-13T12%3A48%3A27Z#L34)
> 
[...]

PR #472, opened 2023-05-17
satellite: remove /usr/lib mount

WanzenBug:
In bd29628, we started bind mounting /usr/lib, as our debian images required it to find the necessary scripts to build DRBD.

A better way to resolve this issue is to include those scripts directly in the container images. This means we can remove the /usr/lib mount, which led to some "franken"-container scenarios.

Depends on: https://github.com/piraeusdatastore/piraeus/pull/145
See also: https://github.com/piraeusdatastore/piraeus-operator/pull/444 and https://github.com/piraeusdatastore/piraeus-operator/issues/427

### DIFF
--- a/pkg/resources/satellite/patches/precompiled-module.yaml
+++ b/pkg/resources/satellite/patches/precompiled-module.yaml
@@ -12,16 +12,11 @@
       volumes:
       - name: usr-src
         $patch: delete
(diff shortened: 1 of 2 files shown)


JoelColledge (approved):
> "franken"-container

:joy:

PR #475, opened 2023-05-23
Support Kernel Build System (Debian 11)

kvaps:
Debian 11 installs the [linux-kbuild](https://packages.debian.org/bullseye/linux-kbuild-5.10) as dependency to [kernel-headers](https://packages.debian.org/buster/linux-headers-5.10.0-0.deb10.19-amd64) package.

The tools and Makefile from it are used directly by `entry.sh` script of kernel-module-injector

The problem is that [most of files](https://packages.debian.org/buster-backports/amd64/linux-kbuild-5.10/filelist) from this package are located under `/usr/lib` directory which we can't directly passtrough into container, because bash and other libs inside the container are using them.

We can't also passtrough the `/usr/lib/linux-kbuild-5.10` directory as the name is unpredictable for it and depending only on kernel version.

Solution: pass it trough under `/host/usr/lib` directory and setup symlinks for non-existing paths into `/usr/lib` of the container


WanzenBug:
Just last week we added `linux-kbuild-*` in the debian containers: https://github.com/piraeusdatastore/piraeus/pull/145

From my tests this seems to be enough to fix building with the existing Operator on Debian.

kvaps:
Okay then this PR is for consideration

I was just thinking if there are no other cases like that, when kernel-headers demands on anything from `/usr/lib`?
If not, we can limit pattern to symlink only them `/usr/lib/linux-kbuild-*`

Another question: Is it correct way to install `linux-kbuild-*` from debian and ship it with the docker image? What if Ubuntu will go the same way by separating kbuild files from kernel-headers, then we will use debian tools to compile ubuntu's kernel-headers, can we be sure that it will work in that way?

WanzenBug:
> I was just thinking if there are no other cases like that, when kernel-headers demands on anything from /usr/lib?

For now, Debian seem to be the only ones doing this micro-packaging, in all other cases /usr/src and /lib/modules was enough

> Another question: Is it correct way to install linux-kbuild-* from debian and ship it with the docker image?

If we use the debian image to build, we need to have the debian linux-kbuild-* tools. The tools depend on shared libraries, and those need to be available in the container. 

If Ubuntu decided they also want to do this split with linux-kbuild-*, and we would link those from the host, I think this would break. Because the tools try to load some shared libraries that don't exist in a Debian container. Luckily, if Ubuntu switches, we would use our Ubuntu containers in any case.

WanzenBug:
This seems to be handled for the most part by adding the linux-kbuild-* packages to the containers. There is also now https://piraeus.io/docs/v2/how-to/raspian/#configure-the-drbd-module-loader-for-raspberry-pi-os-raspian for more esoteric distributions.

PR #479, opened 2023-05-26
CSI: report seLinuxMount capability

WanzenBug:
With this capability set, kubelet can pass "-o context=..." mount options
to automatically relabel all files. Otherwise kubelet will need to run
"chcon -R", which might time out for large volumes without any clear error
indication.
    
Since our CSI driver just passes the options to mount(), we can set this
without any further changes.

Also contains updates to the build tools, cleaning up generated code along the way


WanzenBug:
Note: Doesn't work on Kubernetes 1.24, as the `seLinuxMount` field does not exist. Wait until 1.25 is the old "stable" release.

JoelColledge (approved):
Need to remember to remove all this when we no longer support versions older than 1.25.

PR #484, opened 2023-06-15
use consistent tolerations for all "daemon" pods

WanzenBug:
Use the same tolerations for all 3 kinds of daemon pods:

* Satellite
* CSI Node
* HA Controller

These tolerations allow running on control plane nodes, and also tolerate the HA Controller managed taints.

The satellite pods also include the tolerations for node pressure events, they are included in the normal daemon set pods automatically, but the satellite is a custom pods.


JoelColledge (approved):
I was initially a bit confused because the HA Controller is mentioned in the commit message, but isn't touched. Then I realized that those tolerations were already correct.

PR #485, opened 2023-06-16
Disable operator metrics reporting by default

WanzenBug:
The Operator metrics are not generally useful without additional metrics. They also depend on an external container.

In order to streamline air-gapped deployments, remove this feature from the default deployment.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -23,10 +23,10 @@ resources:
 - ../webhook
 - ../certmanager
 
-patchesStrategicMerge:
-- manager_auth_proxy_patch.yaml
-- manager_webhook_patch.yaml
-- webhookcainjection_patch.yaml
+patches:
+- path: manager_webhook_patch.yaml
+- path: webhookcainjection_patch.yaml
+# - path: manager_auth_proxy_patch.yaml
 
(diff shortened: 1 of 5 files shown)


config/default/kustomization.yaml
```diff
@@ -23,10 +23,10 @@ resources:
 - ../webhook
 - ../certmanager
 
-patchesStrategicMerge:
-- manager_auth_proxy_patch.yaml
-- manager_webhook_patch.yaml
-- webhookcainjection_patch.yaml
+patches:
+- path: manager_webhook_patch.yaml
+- path: webhookcainjection_patch.yaml
```
JoelColledge, line 28: What does `webhookcainjection_patch.yaml` have to do with the operator metrics?

WanzenBug: Nothing. In this case I commented out the `manager_auth_proxy_patch.yaml` patch and upgraded the `patchesStrategicMerge:` section to `patches:`, as newer kustomize versions warn on `patchesStrategicMerge`

JoelColledge: :bulb:

JoelColledge (approved):

PR #486, opened 2023-06-19
Improvements for kmod loading

WanzenBug:
* Better handling of SELinux enabled systems
* Check for module parameters on systems with DRBD already loaded
* Image updates to support the above changes

See commit messages for details

### DIFF
--- a/pkg/resources/satellite/pod/node-pod.yaml
+++ b/pkg/resources/satellite/pod/node-pod.yaml
@@ -14,6 +14,9 @@ spec:
   initContainers:
     - name: drbd-module-loader
       image: drbd-module-loader
+      env:
+        - name: LB_FAIL_IF_USERMODE_HELPER_NOT_DISABLED
+          value: "yes"
       securityContext:
         # NB: We don't use RO root here, as SELinux likes to complain
         # if we try to insert a module from a emptyDir volume.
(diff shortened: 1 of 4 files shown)


pkg/resources/satellite/pod/node-pod.yaml
```diff
@@ -29,6 +37,8 @@ spec:
         - mountPath: /usr/src
           name: usr-src
           readOnly: true
+        - mountPath: /tmp
+          name: tmp
```
JoelColledge, line 41: This means that the module loader and satellite now use the *same* `/tmp`, right? I guess that is fine.

JoelColledge (approved):

pkg/resources/satellite/pod/node-pod.yaml
WanzenBug: Yes. I didn't really think about it, but a separate /tmp may make sense.

On the other hand, from previous tests I know that kubernetes doesn't clean up these emptyDir volumes once they become unused, but only after the whole pod is removed. So while it does provide more separation, it does not disk usage.

JoelColledge: > a separate /tmp may make sense.

Perhaps more logical, but irrelevant in practice, I would say. The module loader does use static paths `/tmp/pkg` and `/tmp/ko`, but the satellite certainly won't conflict with those or be affected by their presence, so we're fine.

PR #490, opened 2023-06-20
generate webhook cert without cert-manager

WanzenBug:
Having a dependency on cert-manager prevents Piraeus from working "out of the box". While the validation webhook is optional, it does improve the user experience, so we want to keep it.

The solution is to have our own self-contained provisioning service for the webhook. This is deployed by default and ensures we have a valid TLS secret and up-to-date validation webhook config.

### DIFF
--- a/cmd/gencert/gencert.go
+++ b/cmd/gencert/gencert.go
@@ -1,0 +1,282 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"crypto/x509"
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/cert"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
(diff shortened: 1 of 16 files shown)


cmd/gencert/gencert.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (g *GenCertRunner) Start(ctx context.Context) error {
+	g.Logger.Info("starting cert renew")
+	for {
+		c, err := g.reconcile(ctx)
+		if err != nil {
+			return err
+		}
+
+		renewIn := g.needsRenewIn(c, 15*24*time.Hour)
```
JoelColledge, line 154: I hope we have a test for this :wink:

Just leave the test runner going for 2 weeks...

cmd/gencert/gencert.go
```diff
@@ -0,0 +1,281 @@
...
+	if !sets.NewString(c.DNSNames...).Equal(sets.NewString(g.dnsNames()...)) {
+		return 0
+	}
+
+	now := time.Now()
+	if now.Before(c.NotBefore) {
+		return 0
+	}
+
+	return c.NotAfter.Sub(now.Add(renewBefore))
```
JoelColledge, line 273: What happens if `now < NotBefore` and `NotAfter < now + renewBefore`? Can only happen if we can have `NotAfter - NotBefore < renewBefore`. That is, a certificate that has a narrow renewal window. Won't this send our reconciler into a busy loop?

JoelColledge (commented):
Code looks good and plausible. It would be good to include some explanation somewhere describing what this is. So a user understands why they have this Deployment. And why `cert-manager` is still required for other purposes. Perhaps worth adding a "explanation" doc listing the different components and how they interrelate?

cmd/gencert/gencert.go
WanzenBug: It should be fine, since the self-signed certs are valid for a year.

cmd/gencert/gencert.go
WanzenBug: We have now :)

docs/explanation/components.md
```diff
@@ -0,0 +1,76 @@
+# Understanding Piraeus Datastore Components
+
+Piraeus Datastore consists of seven different components. Every component is running as a separate Deployment, DaemonSet or plain Pod.
```
JoelColledge, line 3:
```suggestion
Piraeus Datastore consists of several different components. Each component runs as a separate Deployment, DaemonSet or plain Pod.
```

A number like "seven" is unlikely to be updated when the list is changed.

docs/explanation/components.md
```diff
@@ -0,0 +1,76 @@
...
+It provides an external API used by LINSTOR CSI and Piraeus Operator to change the cluster state.
+
+# `linstor-satellite`
+
+The LINSTOR Satellite service runs on each node. It acts as the local configuration agent for LINSTOR managed storage.
+It is stateless and receives all the information it needs from the LINSTOR Controller.
+
+# `linstor-csi-controller`
+
+The LINSTOR CSI Controller Pod creates, modifies, deletes volumes and snapshots. It translates the state of Kubernetes
```
JoelColledge, line 59:
```suggestion
The LINSTOR CSI Controller Pod creates, modifies and deletes volumes and snapshots. It translates the state of Kubernetes
```

JoelColledge (commented):
Wonderful! I think we can remove this note from the readme:

> We are currently working on improving our documentation. Additional tutorials and explanation will be added soon.

docs/explanation/components.md
```diff
@@ -0,0 +1,76 @@
...
+The LINSTOR Controller is responsible for resource placement, resource configuration, and orchestration of any
+operational processes that require a view of the whole cluster. It maintains a database of all the configuration
+information for the whole cluster, stored as Kubernetes objects.
+
+The LINSTOR Controller connects to the LINSTOR Satellites and sends them instructions for achieving the desired cluster
+state.
+
+It provides an external API used by LINSTOR CSI and Piraeus Operator to change the cluster state.
+
+# `linstor-satellite`
```
JoelColledge, line 52: Perhaps mention why the satellites are created as plain pods? It is unusual.

WanzenBug: Done

JoelColledge (approved):

PR #500, opened 2023-07-11
small fixes in preparation for node connection support

WanzenBug:
Adds a few small fixes in preparation for node connection support.

* Update the PROJECT file to reflect the current state of the project. This does not seem to be used for anything important, but new resources will populate the file, so we fill in the information about existing resources.
* Update the LINSTOR property handling not require a "metadata property" for empty property lists. Previously the metadata property was always applied, even for an empty "[]" value.


pkg/linstorhelper/properties.go
```diff
@@ -46,6 +46,10 @@ func MakePropertiesModification(current, expected map[string]string) *lclient.Ge
 		}
 	}
 
+	if len(lastApplied) > 0 && len(expected) == 0 {
```
JoelColledge, line 49: Why only when `len(lastApplied) > 0`? If we're deleting all the properties we have set, then we can delete `LastApplyProperty` immediately as well, right?

WanzenBug: True, don't really know why this check existed.

JoelColledge (approved):

PR #508, opened 2023-07-27
Upgrade HA Controller from v0.3 to v1

wangjiusi:
The HA Controller v1 is currently deployed separately from Operator v1. This PR merges them two. 
Special considerations are made as below: 

1. ha-controller chat is downloaded directory from its helm chart repository, using the latest version; 
2. ha-controller chat is added as an optional dependency;
3. `yq` is used to modify names and update version numbers;
4. the name 'ha-controller' is used, instead of 'haController', because the chart name is used in the chart for service-account name, which disallows any capitalized letter;


WanzenBug (changes requested):
Could we not simply specify?
```
dependencies: 
  - name: ha-controller
    version: "1.1.4"
    repository: https://piraeus.io/helm-charts/
    condition: haController.enabled
```

alexzhc:
> Could we not simply specify?
> 
> ```
> dependencies: 
>   - name: ha-controller
>     version: "1.1.4"
>     repository: https://piraeus.io/helm-charts/
>     condition: haController.enabled
> ```

We thought about it, but would prefer keeping ha-controller charts in the repo for the sake of convenience, especially in the air-gapped situation.

WanzenBug:
> We thought about it, but would prefer keeping ha-controller charts in the repo for the sake of convenience, especially in the air-gapped situation.

As a maintainer I don't find it very convenient to have to update the chart in two places. I think users can be expected to run `helm dependency build charts/piraeus` once in the case of air-gapped deployments.

alexzhc:
@WanzenBug  I agree with your approach and have tried it out. There is only a minor issue with the naming: it seems `nameOverride` does not work on dependency charts, so we could get a name such as `piraeus-piraeus-ha-controller`. It there a way to avoid it?

```
$ helm upgrade --install --namespace piraeus-system --create-namespace piraeus ./charts/piraeus \
--set csi.kubeletPath=/var/snap/microk8s/common/var/lib/kubelet/ \
--set haController.enabled=true \
--set haController.nameOverride="ha-controller"

$ kubectl -n piraeus-system get po
NAME                                      READY
piraeus-etcd-0                            0/1  
piraeus-operator-6656896cd4-fx2p7         1/1  
piraeus-csi-node-cdg2z                    0/3  
piraeus-csi-node-28lgz                    0/3  
piraeus-csi-node-vrpcc                    0/3  
piraeus-csi-node-xhgqn                    0/3  
piraeus-csi-controller-64cd4ccc75-2z4qx   0/6  
piraeus-piraeus-ha-controller-bnx97       1/1  
piraeus-piraeus-ha-controller-5j6fj       1/1  
piraeus-piraeus-ha-controller-v9nw9       1/1  
piraeus-piraeus-ha-controller-gk8v7       1/1  
piraeus-ns-node-xq8r8                     0/2  
piraeus-ns-node-qknfj                     0/2  
piraeus-ns-node-hczb8                     0/2  
piraeus-ns-node-bckbj                     0/2  
piraeus-cs-controller-6db95bb964-drlc9    1/1  

```

charts/piraeus/Chart.yaml
```diff
@@ -9,3 +9,7 @@ dependencies:
   - name: "etcd"
     version: "0.7.5"
     condition: etcd.enabled
+  - name: "piraeus-ha-controller"
+    version: "1.1.4"
+    repository: "https://piraeus.io/helm-charts/"
+    condition: ha-controller.enabled
```
WanzenBug, line 15: Should be `piraeus-ha-controller.enabled`.

charts/piraeus/values.cn.yaml
```diff
@@ -123,12 +123,15 @@ operator:
...
+  imagePullSecrets: []
+  nameOverride: "ha-controller"
+  fullnameOverride: ""
   resources: {}
-  replicas: 1
-  customLabels: {}
-  customAnnotations: {}
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
```
WanzenBug, line 137: I understand why we need to override the repository, but why do we need to specify the other values too?

charts/piraeus/values.yaml
```diff
@@ -123,12 +123,15 @@ operator:
...
+  imagePullSecrets: []
+  nameOverride: "ha-controller"
+  fullnameOverride: ""
   resources: {}
-  replicas: 1
-  customLabels: {}
-  customAnnotations: {}
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
```
WanzenBug, line 137: Same as above regarding the values, but in this case we do not even need the repository, I think?

WanzenBug (changes requested):

charts/piraeus/Chart.yaml
alexzhc: corrected.

charts/piraeus/values.cn.yaml
alexzhc: yes, I leave there only `resources, affinity and tolerations` now

charts/piraeus/values.yaml
alexzhc: yes, I leave there only `resources, affinity and tolerations` now

alexzhc:
@WanzenBug we have updated per your review, please have a look, thanks.

WanzenBug (approved):

PR #509, opened 2023-07-27
satellite: restart terminated pods

WanzenBug:
We expect satellite to be recreated when they shut down for any reason, as long as they are still on a matching node.

This was not always the case, as the Pod could get into a terminated state, which even "restartPolicy: Always" did not cover. This may happen when the Pod is shut down forcefully during node shutdown.

This patch modifies the PodPatcher to also consider the phase of the old Pod when deciding if a delete+create operation is necessary.


WanzenBug:
@kvaps would be great if you could verify that this fixes the issue?

JoelColledge (approved):

PR #523, opened 2023-08-31
Support complex node affinity in cluster spec

### LINKED ISSUE #522: Adding support "nodeAffinity" to LinstorCluster config
>  If it was supporting nodeAffinity, you'd be able to exclude..

Please open a feature request, should not be too hard to implement :)

_Originally posted by @WanzenBug in https://github.com/piraeusdatastore/piraeus-operator/issues/519#issuecomment-1698584288_

WanzenBug:
Use the nodeAffinity, like one would set it on pods to allow more complex deployment scenarios. We need to respect these settings when creating Satellites, but also when creating our own deployments.

Fixes #522


docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -27,6 +27,33 @@ spec:
       value: "no"
 ```
 
+### `.spec.nodeAffinity`
+
+Selects which nodes the LinstorSatelliteConfiguration should apply to. If empty, the configuration applies to all nodes.
+
+When this is used together with `.spec.nodeAffinity`, both need to match in order for the configuration to apply to a
```
JoelColledge, line 34:
```suggestion
When this is used together with `.spec.nodeSelector`, both need to match in order for the configuration to apply to a
```

JoelColledge:
The rest all looks good :+1:

docs/reference/linstorsatelliteconfiguration.md
WanzenBug: You've got :eagle: :eyes: 

:smile:

JoelColledge (approved):

PR #552, opened 2023-10-30
satellites: add TLS handshake daemon for DRBD

WanzenBug:
Adds a new switch .spec.internalTLS.tlsHandshakeDaemon on the LinstorSatelliteConfiguration. If enabled, it configures another sidecar on the Satellite that runs "tlshd" with the same key material as the LINSTOR Satellite.

With this enabled, users can set "DrbdOptions/Net/tls: yes" on their storage class to enable TLS encryption for DRBD connections.


api/v1/tls.go
```diff
@@ -14,3 +14,14 @@ type TLSConfig struct {
 	//+kubebuilder:validation:Optional
 	CertManager *cmmetav1.ObjectReference `json:"certManager,omitempty"`
 }
+
+type TLSConfigWithHandshakeDaemon struct {
+	TLSConfig `json:",inline"`
+
+	// TLSHandshakeDaemon enables tlshd for establishing TLS session for use by DRBD.
```
JoelColledge, line 21:
```suggestion
	// TLSHandshakeDaemon enables tlshd for establishing TLS sessions for use by DRBD.
```

api/v1/tls.go
```diff
@@ -14,3 +14,14 @@ type TLSConfig struct {
...
+	//
+	// If enables, adds a new sidecar to the LINSTOR Satellite that runs the tlshd handshake daemon.
```
JoelColledge, line 23:
```suggestion
	// If enabled, adds a new sidecar to the LINSTOR Satellite that runs the tlshd handshake daemon.
```

api/v1/tls.go
```diff
@@ -14,3 +14,14 @@ type TLSConfig struct {
...
+	// The daemon uses the tls certificate and key to establish secure connections on behalf of DRBD.
```
JoelColledge, line 24:
```suggestion
	// The daemon uses the TLS certificate and key to establish secure connections on behalf of DRBD.
```

JoelColledge (commented):
Add a test with `tlsHandshakeDaemon: true`?

JoelColledge (approved):

PR #561, opened 2023-11-10
deploy: add HA Controller taints to Operators' tolerations

WanzenBug:
If the whole cluster reboots, it is very likely that all cluster nodes get tainted by the HA Controller. Because the Operator does not start, it can't get the satellites up and running, and we can never clear the taints.

The fix is to include the tolerations in the default deployment for the operator (and gencert, too).


JoelColledge (approved):
Sounds like something that was learned by experience :smile:

PR #570, opened 2023-11-29
Adds resource options to directly change the pod template

WanzenBug:
Most modifications users will want to make is centered on the pod templates used. In particular, they may want to set resource requests and limit, or configure some special networking option.

Using .spec.patches for that is cumbersome for most use cases, so we add a short cut in the form of `podTemplate` parameters,
which take a (skeleton) of a Pod's definition. The templates are converted to a regular patch and applied to our built-in resources.

We also add the option to disable some components: a user might want to disable the HA Controller or have their own CSI Controller deployment for whatever reason.


api/v1/linstorcluster_types.go
```diff
@@ -86,6 +86,22 @@ type LinstorClusterSpec struct {
 	// + See LinstorSatelliteSpec.InternalTLS for why nullable is needed.
 	// +nullable
 	ApiTLS *LinstorClusterApiTLS `json:"apiTLS,omitempty"`
+
+	// Controller controls the deployment of the LINSTOR Controller Deployment.
+	// +kubebuilder:validation:Optional
+	Controller *ComponentSpec `json:"controller,omitempty"`
+
+	// Controller controls the deployment of the CSI Controller Deployment.
```
JoelColledge, line 94:
```suggestion
	// CSIController controls the deployment of the CSI Controller Deployment.
```

pkg/merge/linstorsatelliteconfiguration.go
```diff
@@ -42,6 +44,25 @@ func SatelliteConfigurations(node *corev1.Node, configs ...piraeusv1.LinstorSate
...
 
+		if len(cfg.Spec.PodTemplate) > 0 {
+			var u unstructured.Unstructured
+			_ = json.Unmarshal(cfg.Spec.PodTemplate, &u)
+			u.SetAPIVersion("v1")
+			u.SetKind("Pod")
+			u.SetName("satellite")
+
+			encoded, err := u.MarshalJSON()
+			if err == nil {
```
JoelColledge, line 55: Worthwhile at least logging an error if `err != nil`? Can't imagine why it would occur, though.

JoelColledge (commented):
I notice that you build one typo into each PR, just to make sure I'm actually reading them :wink: 

Looks clean and reasonable to me.

PR #572, opened 2023-12-01
chart improvements

WanzenBug:
* Don't upgrade from v1
* Allow image config overides during deployment
* Include node connections webhook

### DIFF
--- a/charts/piraeus/values.yaml
+++ b/charts/piraeus/values.yaml
@@ -97,3 +97,12 @@ podDisruptionBudget:
   enabled: true
   minAvailable: 1
   # maxUnavailable: 1
+
+imageConfigOverride: []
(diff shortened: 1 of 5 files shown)


charts/piraeus/values.yaml
```diff
@@ -97,3 +97,12 @@ podDisruptionBudget:
   enabled: true
   minAvailable: 1
   # maxUnavailable: 1
+
+imageConfigOverride: []
+  # - base: example.com/piraeus
+  #   components:
+  #     linstor-csi
```
JoelColledge, line 104:
```suggestion
  #     linstor-csi:
```

To make the yaml valid, right?

JoelColledge (approved):

PR #574, opened 2023-12-05
Release 2.3.0

WanzenBug:
Apart from the release commits, this also changes the way release script so that the instructions on the main branch always point to the latest released version. This should help users when deploying using the default instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [v2.3.0] - 2023-12-05
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


JoelColledge:
This basically removes the rolling release option. The "dev cycle" now only exists in the changelog. There is no absolute requirement for supporting rolling release deployment. I'm just wondering what has changed since https://github.com/piraeusdatastore/piraeus-operator/pull/411? That time we ended up keeping the option.

WanzenBug:
Turns out that people mostly copy+paste whatever is on the front page. Which leads to people deploying the "rolling release". Which is mostly fine to get startet, since it all starts up fine.

But users may run into issues if the operator image needs to be pulled again at a later time, for example because of a Kubernetes upgrade. Then, the deployment for the Operator is old, the image is new and you may run into issues.

I changed it so that is you really want to, you can still do "rolling" release with `ref=v2`, but I no longer want it as the default approach.

JoelColledge:
Sounds like a good solution. It makes sense to leave the documentation on the latest release. We no longer have a rolling version of the Helm chart, but I don't think that is very important.

JoelColledge (approved):

PR #586, opened 2024-01-04
Use DaemonSet to deploy satellites

WanzenBug:
With this PR, we start using daemonsets for deploying satellite pods. Like before, there is only one actual Pod controlled by a specifc `LinstorSatellite` instance. We make this change so that:
* Using `kubectl drain ...` works as expected, leaving satellites untouched
* We no longer need manually manage the pod lifecycle, the daemonset will take care of restarting and updating Pods as appropriate

We can now do this, since LINSTOR learned to refresh it's internal cache on reconnect: during the reconnect, it will re-read the `uname -n` and update all resources with the new name. That means we no longer need to have a stable hostname for the containers themselves. If a user is using the host network, this obviously was never an issue, as the host name is stable in this case.

Most of the work for this change is making sure that existing configuration by the users still apply. This is mostly an issue with user provided `patches`, that likely refer to the `satellite` Pod. These patches are transparently converted to patches for the daemon set, so users will not see the change.

To aid migration, there is also a new warning when a patch refers to the bare `satellite` Pod: users get encouraged to patch the daemonset instead.

There is also a lot of documentation that needed to be updated. This is now using `podTemplate`, as that is both simpler to configure and independent of the actual resource deployment by the operator.

For review, I recommend reviewing commits one by one.


JoelColledge:
I'm just starting to look at this and am thinking about the concept. In particular, what it means for debugging. Suppose the satellite pod on node A is deleted and starts again. The pod name will change, right? And thus the hostname in the container. On node B the connections will be renamed. So in the kernel log for B we will see "connection lost to linstor-satellite-A-foooo" followed by "connection established to linstor-satellite-A-baaar". At least it is just to ending that changes. Still going to take a bit of getting used to.

I'm tempted to suggest we un-deprecate connection naming and use it to reduce the connection names back to the host hostname...

WanzenBug:
> I'm tempted to suggest we un-deprecate connection naming and use it to reduce the connection names back to the host hostname...

Another solution (suggested by kvaps, I believe) was to manually set the hostname on start up. In theory, this could be done with an init container, and I initially planned on doing that, but I could not find the right snippet that does that. Then I noticed that LINSTOR actually can deal with changing hostnames now, so I skipped it.

It would also introduce some conflict then using `hostNetwork: true`, as we probably should not mess with the host net namespace.

Another option I guess would be to support overriding the kernel hostname with an environment variable in drbd-utils. I saw there is already `__DRBD_NODE_NAME__` used for tests, but that probably shouldn't be used.

Anyways, this is not a pressing issue, so I'd keep this issue open until we have a satisfying solution.

pkg/barepodpatch/barepodpatch.go
```diff
@@ -0,0 +1,115 @@
...
+	"sigs.k8s.io/yaml"
+
+	piraeusiov1 "github.com/piraeusdatastore/piraeus-operator/v2/api/v1"
+	"github.com/piraeusdatastore/piraeus-operator/v2/pkg/utils"
+)
+
+// ConvertBarePodPatch converts a patch for a bare "satellite" pod into a patch for a "linstor-satellite" daemonset
+//
+// Bare pod patches are recognized by the detected target using "kind: Pod" and "name: satellite". It converts
+// both strategic merge patches
```
JoelColledge, line 13: Sentence seems unfinished.

pkg/barepodpatch/barepodpatch.go
```diff
@@ -0,0 +1,115 @@
...
+		if target.Kind != "Pod" || target.Name != "satellite" {
+			continue
+		}
+
+		if convertedPatch := convertStrategicMergePatch([]byte(patches[i].Patch)); len(convertedPatch) > 0 {
+			patches[i].Patch = string(convertedPatch)
+		} else if convertedPatch := convertJsonPatch([]byte(patches[i].Patch)); len(convertedPatch) > 0 {
+			patches[i].Patch = string(convertedPatch)
+		} else {
+			continue
```
JoelColledge, line 30: This is an error condition, but I guess in this case we will try to apply the unconverted patch and get an error then, so that is good enough.

JoelColledge (commented):
Looks great. Very clean. Appropriate test coverage.

JoelColledge (approved):

JoelColledge:
In principle, I think we could have DRBD use the host hostnames without any of these workarounds:

> Another solution (suggested by kvaps, I believe) was to manually set the hostname on start up.

> Another option I guess would be to support overriding the kernel hostname with an environment variable in drbd-utils. I saw there is already `__DRBD_NODE_NAME__` used for tests, but that probably shouldn't be used.

> un-deprecate connection naming

All we need is for LINSTOR to write DRBD res files that use the pod hostname for the local node and the host hostnames for the remote nodes. The local node name never reaches DRBD. It is only used by `drbdadm` to find which config stanza to apply.

That would require LINSTOR to understand this separation of names. I'm not sure whether the additional complexity is worth it.

PR #595, opened 2024-01-29
add webhook for validating storage classes

WanzenBug:
Add a webhook validating the parameters section of a Piraeus StorageClass.

With the move away from using "volumeBindingMode: Immediate" to "WaitForFirstConsumer" on storage classes, users often did not see the exact error generated by invalid storage class parameters: They only saw a message by the kubernetes schedule, indicating that there is not enough free space on any node.

To prevent this, and make wrong storage class parameters observable in general, we add a new webhook that validates all storage classes using the LINSTOR CSI provisioner. We simply include the volume parameter parsing logic from the latest LINSTOR CSI release.


JoelColledge (commented):
It seems to me that it would make more sense to make this part of the LINSTOR CSI project. Would it be possible to host it in the LINSTOR CSI Controller?

WanzenBug:
> Would it be possible to host it in the LINSTOR CSI Controller?

Sure, and I agree that it makes sense there. I just put it in here as we have all the necessary infrastructure already wired up here. For Webhooks, we need a `Service`, a `ValidatingWebhookConfiguration` and crucially, a TLS configuration. That last one is the one that makes this set up complicated. We don't want to require cert-manager for the deployment, but it also feels wrong to add the complexity of `gencert` to the csi-controller deployment...

Maybe there is some sort of middle ground where we can reuse the basic cert-provisioning logic for the csi-controller.

JoelColledge (approved):
Good point. We could just see this as an extra bit of functionality provided by the operator.

PR #603, opened 2024-02-15
Satellite daemonset fixes

WanzenBug:
* Also on control-plane nodes: this is to stay compatible with existing deployments.
* Force `uname -n` to return the "root" node name. This requires the change in https://github.com/piraeusdatastore/piraeus/pull/172 with a longer explanation on why and how.

### DIFF
--- a/pkg/resources/satellite/satellite/daemonset.yaml
+++ b/pkg/resources/satellite/satellite/daemonset.yaml
@@ -69,6 +69,10 @@ spec:
           env:
             - name: JAVA_OPTS
               value: -Djdk.tls.acknowledgeCloseNotify=true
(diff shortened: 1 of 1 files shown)


JoelColledge (approved):

JoelColledge:
@WanzenBug Perhaps we should add some "explanation" docs for the network namespaces that we use? We now have host, pod and satellite-namespace levels.

While about it, I noticed that we have this statement, which is no longer valid, right?

> This also means that unless [`host-networking](https://github.com/piraeusdatastore/piraeus-operator/blob/v2/docs/how-to/drbd-host-networking.md) is used, the DRBD connection names will use the Pod hostnames instead of the node name.

PR #609, opened 2024-02-27
Add documentation for using Piraeus behind a HTTP proxy

simonbjorzen-ts:


docs/how-to/http-proxy.md
```diff
@@ -0,0 +1,31 @@
+# How to run Piraeus when using a HTTP proxy (like Squid)
```
WanzenBug, line 1:
```suggestion
# How to Use Piraeus Datastore with a Cluster Behind HTTP Proxy
```

docs/how-to/http-proxy.md
```diff
@@ -0,0 +1,31 @@
...
+
+When attempting to deploy Piraeus behind a HTTP proxy, the following configuration needs to be done.
```
WanzenBug, line 3: Our guides usually start with a small introduction like:

> This guide shows you how to ... 
> 
> To complete this guide, you should be familiar with: ...

See https://github.com/piraeusdatastore/piraeus-operator/blob/v2/docs/how-to/talos.md for examples.

docs/how-to/http-proxy.md
```diff
@@ -0,0 +1,31 @@
...
+
+```yaml
+apiVersion: piraeus.io/v1
+kind: LinstorSatelliteConfiguration
+metadata:
+  name: custom-settings # Set a name of the configuration
```
WanzenBug, line 9:
```suggestion
  name: http-proxy
```

Give this a name that describes what it actually does.

docs/how-to/http-proxy.md
```diff
@@ -0,0 +1,31 @@
...
+    spec:
+      containers:
+        - name: linstor-satellite
+          env:
+            - name: HTTP_PROXY
+              value: http(s)://proxy-ip:port
+            - name: HTTPS_PROXY
+              value: http(s)://proxy-ip:port
+            - name: NO_PROXY
+              value: localhost,127.0.0.1,10.0.0.0/8,172.16.0.0/12 # Add internal IP ranges and domains here
```
WanzenBug, line 21: Instead of having placeholders here, use concrete values and describe them in the paragraph before the code block. So something like:

> The following configuration configures an HTTP Proxy reachable at `https://proxy.example.com:8080`:
>
> ```yaml
> apiVersion: piraeus/v1
> ...
> ```

docs/how-to/http-proxy.md
```diff
@@ -0,0 +1,31 @@
...
+      initContainers:
+        - name: drbd-module-loader
+          env:
+            - name: HTTP_PROXY
+              value: http(s)://proxy-ip:port
+            - name: HTTPS_PROXY
+              value: http(s)://proxy-ip:port
+            - name: NO_PROXY
+              value: localhost,127.0.0.1,10.0.0.0/8,172.16.0.0/12 # Add internal IP ranges and domains here
+```
```
WanzenBug, line 31: Maybe add an explanation why each section is needed? I see the need for the `drbd-module-loader`, because that may call out to `spaas.drbd.io`, but why does the `linstor-satellite` need it?

WanzenBug (changes requested):
Thank you for the contribution. I have left some suggestions.

You also need to sign off your commit. See https://github.com/piraeusdatastore/piraeus-operator/pull/609/checks

docs/how-to/http-proxy.md
simonbjorzen-ts: Changed accordingly

WanzenBug (dismissed):

WanzenBug (approved):

PR #611, opened 2024-03-01
ci: run basic tests for promised lower compat version

WanzenBug:
We try to support older kubernetes versions: to ensure that they actually work, we run our normal tests against the oldest kubernetes version we support.

This already managed to unearth some issues with certain patches being applied.


JoelColledge (approved):
Is `make test` actually run in CI?

WanzenBug:
> Is make test actually run in CI?

Yes, it's part of the pre-commit pipeline. I specifically did not add the `compat-test` there, as the normal tests already take long enough.

PR #633, opened 2024-03-28
satellites: add support for ZFS Pools

WanzenBug:
Add support for ZFS pools, adding to the existing types of storage pools. This is mostly calling the right API in LINSTOR.

One tricky part is that "zpool create" called when using LINSTOR to create the zpool will wait for a specific file to appear in /run/udev, so we need to bind mount that from the host.

I've verified that this directory is present in all distributions we have in our "how-to" section, so I do not exepct any issues there.


docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -122,7 +126,10 @@ This example configures three LINSTOR Storage Pools on all satellites:
   devices `/dev/sdb` and `/dev/sdc` if it does not exist already.
 * A File System Pool named `fs1`. It will use the `/var/lib/linstor-pools/fs1` directory on the host, creating the
   directory if necessary.
-* A File System Pool namef `fs2`, using sparse files. It will use the custom `/mnt/data` directory on the host.
+* A File System Pool named `fs2`, using sparse files. It will use the custom `/mnt/data` directory on the host.
```
JoelColledge, line 129: > This example configures *three* LINSTOR Storage Pools on all satellites:

Always risky to write the length of a list :smile: Replace with "these"?

JoelColledge (commented):
Just the one nit. Otherwise all looks good.

JoelColledge (approved):

PR #652, opened 2024-04-23
add client-side caching to LINSTOR client

WanzenBug:
On every reconciliation we run a create some LINSTOR requests. Most of them are "read-only", to ensure the cluster is in the expected state. These requests often come in "batches", i.e. update the status of all nodes.
    
With a new feature of the golinstor client, we can cache these respones and only use a single request for _all_ nodes or _all_ storage pools. This drastically reduces the number of requests in clusters with many nodes, or clusters with fast-changing labels.
    
The cache resets on every "modification" made by the client, ensuring the cluster state is at least internally consistent.
    
One tricky part is that we actually create lots of new clients for every reconciliation. These should still share the same cache, provided they actually point at the same cluster. This is done by inserting a another level of indirection by creating the cache once per controller URL.

### DIFF
--- a/pkg/linstorhelper/client.go
+++ b/pkg/linstorhelper/client.go
@@ -8,8 +8,12 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
+	"time"
 
+	lapicache "github.com/LINBIT/golinstor/cache"
 	lapi "github.com/LINBIT/golinstor/client"
+	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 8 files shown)


pkg/linstorhelper/client.go
```diff
@@ -18,11 +19,26 @@ import (
...
+
 // Client is a LINSTOR client with convenience functions.
 type Client struct {
 	lapi.Client
 }
 
+// PerClusterOption makes an option that is shared between all instances connecting to the same cluster.
+// This can be used to make a cached shared between different client instances or a rate limiter applied to all
+// clients connected to the same controller.
+func PerClusterOption(opt func() lapi.Option) lapi.Option {
```
JoelColledge: This is a function which returns a function and has an argument that is a function that returns a function 😵

pkg/linstorhelper/client.go
```diff
@@ -18,11 +19,26 @@ import (
...
+	return func(c *lapi.Client) error {
+		clusterMap, _ := optsCache.LoadOrStore(c.BaseURL(), &sync.Map{})
+		clusterOpt, _ := clusterMap.(*sync.Map).LoadOrStore(&opt, opt())
```
JoelColledge: Here we use a pointer to a function as a map key. This seems rather fragile. This could run twice with `opt` being 2 different functions containing identical code and the behavior would be different from when it runs twice with `opt` being the same function.

JoelColledge (commented):
This level of abstraction makes it hard to understand. And I'm writing as someone who likes Haskell :smile: Perhaps replace `PerClusterOption` with `PerClusterNodeCache`? `PerClusterNodeCache` would basically be the same as `PerClusterOption`, but with `opt` inlined.

That would remove need for nested maps and reduce the number of levels of function indirection.

Does this work?

WanzenBug:
Simplified, as per your suggestion. Not sure why I was so focused on having it "generic" in the first place.

JoelColledge (approved):

PR #657, opened 2024-05-02
add default priority class name

### LINKED ISSUE #654: [bug] The priorityClass of the pods is not set. Cascade of failure ensues.
The `PriorityClass` of all pods created by the default install method is not set. Nor is there any option and/or documentation in the chart/manifests to set it. I therefore wrongly assumed that it was set,

Today a cluster wend down because some disk pressure caused the linstor-satellite pods to be evicted on a couple of nodes. The result is a cascade of failures that is hard to recover from. Luckily, this happened in a staging cluster so nobody is asking to throw Linstor out of the window yet so we can keep testing and validating.

The priorityClass of all pods created by the operator should be set to `system-node-critical`regardless of if it runs in `kube-system` or not. upstream supports this [for a while now](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/). There is no scenario in which this is not preferable

discuss/

WanzenBug:
Priority Classes are used by Kubernetes to control preemption of Pods in resource starvation scenarios. Since Piraeus Datastore is most likely an important part of the cluster, and is needed for other Pods to be scheduled, assign it the system-{node,cluster}-critical priority class.

Fixes #654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Default PriorityClasses for all components
+
 ## [v2.5.0] - 2024-04-03
 
 ### Added
--- a/charts/piraeus/templates/deployment.yaml
+++ b/charts/piraeus/templates/deployment.yaml
@@ -92,6 +92,7 @@ spec:
         runAsNonRoot: true
       serviceAccountName: {{ include "piraeus-operator.serviceAccountName" . }}
       terminationGracePeriodSeconds: 10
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Default PriorityClasses for all components
```
bernardgut, line 12: Note: Requires Kubernetes v1.14+ [stable]

WanzenBug: Should not be an issue, we are already only promise support for v1.20 +

JoelColledge (approved):

bernardgut:
thanks for this

PR #666, opened 2024-05-21
tls: allow custom CA sources

WanzenBug:
Allow users to configure a custom CA to validate all internal certificates. This is particularly useful when using cert-manager with a provisioner which does not add the default "ca.crt" to the certificate secret.

Somehow, this change ended up with a lot more changes then I anticipated. This all started out with a report that using the "ACME provisioner" in cert-manager does not work for our cert manager integration. That provisioner does not include the necessary `ca.crt` key in the secret. So in this case, we need to get the CA information from somewhere else.

From what I can tell, this is generally managed by having a "trust-root" secret or config map, which just contains the CA certificate, and nothing else. In order to support that, we need to merge this information into our normal directory layout of `tls.crt`, `tls.key` and `ca.crt`. Luckily, this can be done with volume projections.

We also need to ensure the operator itself respects this setting, otherwise it will also reject connections to LINSTOR.


api/v1/tls.go
```diff
@@ -25,3 +35,91 @@ type TLSConfigWithHandshakeDaemon struct {
...
+	Kind string `json:"kind,omitempty"`
+	// Name of the resource containing the CA Certificate.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// Key to select in the resource.
+	// Defaults to ca.crt if not specified.
+	// +kubebuilder:default:=ca.crt
+	// +kubebuilder:validation:Optional
+	Key string `json:"key,omitempty"`
+	// Optional specified wheter the resource or its key must be defined
```
JoelColledge:
```suggestion
	// Optional specifies whether the resource and its key must be defined.
```

internal/controller/patches_test.go
```diff
@@ -185,13 +185,13 @@ func TestPatches(t *testing.T) {
...
 			call: func() ([]kusttypes.Patch, error) {
-				return controller.ClusterCSIControllerApiTLSPatch("controller")
+				return controller.ClusterCSIControllerApiTLSPatch("controller", nil)
 			},
 		},
 		{
 			name: "ClusterCSINodeApiTLSPatch",
 			call: func() ([]kusttypes.Patch, error) {
-				return controller.ClusterCSINodeApiTLSPatch("node")
+				return controller.ClusterCSINodeApiTLSPatch("node", nil)
```
JoelColledge: Is it worth adding a test where the argument is not `nil`?

JoelColledge (commented):
All very neat and tidy :+1:

JoelColledge (approved):

PR #668, opened 2024-05-23
ci: fix loading of old envtest assets

WanzenBug:
envtest is moving away from GCS buckets for storage and removed the 1.20 assets from the replacement. Get back the original assets by enabling the deprecated GCS bucket on the setup-envtest binary when installing for compat tests.

Also, cache the binaries in CI to be at least a bit more independant from the external sources.


JoelColledge (approved):
I'm guessing the deprecated assets will disappear at some point and we will have to update before/when that happens.

PR #684, opened 2024-07-17
Release 2.5.2

WanzenBug:


WanzenBug:
@rck usually these are handled by Joel, but since he is on vacation please review. This is mostly just updating images and other release tasks.

The first commit needs a bit of explanation: We watch all resources we create for modification: if we detect modification we run our reconciliation loop again. This is the usual operator pattern. Turns out that on some systems the ServiceAccount resource gets modified automatically, which triggers our loop, which resets the resource, triggering the other component again. So to prevent an endless loop, we specifically filter for only create/delete events.

In an ideal world this filtering would not be necessary, as the Kubernetes API can correctly merge our applied ServiceAccount with whatever modification was made, but I guess whatever does the modification is doing it in a way that k8s gets confused.

rck (approved):
LGTM

Issue #692, opened 2024-08-14
Talos Drbd module


dragon2611:
I think the docs for talos need updating as it seems installing extensions via machine config is deprecated.

You need to build/deploy install image with DRBD support at https://factory.talos.dev/

The section
```
machine:
  install:
    extensions:
      - image: ghcr.io/siderolabs/drbd:9.2.0-v1.3.6
  kernel:
    modules:
      - name: drbd
        parameters:
          - usermode_helper=disabled
      - name: drbd_transport_tcp
      
```
   becomes
   
```
machine:
  kernel:
    modules:
      - name: drbd
        parameters:
          - usermode_helper=disabled
      - name: drbd_transport_tcp
```

As you still need to set the Kernel module parameter, or you won't get any volumes created.

PR #693, opened 2024-08-14
Update docs for talos due to extension changes talos side.

dragon2611:
Removes the image from machine config as this method is deprecated on the talos side, they now expect you to build an image with the components you need and install that way (DRBD in our case)

https://github.com/piraeusdatastore/piraeus-operator/issues/692


WanzenBug (approved):
LGTM

PR #699, opened 2024-08-26
add wildcard properties to satellite spec

WanzenBug:
When setting properties inherited from the Kubernetes Node, it might be useful to copy all labels/annotations matching a specific prefix. To implement this, extend the existing "nodeFieldRef" syntax by allowing the key to contain a "*" character.

When a "\*" character is used, it acts like in a glob pattern, matching all keys on the referenced object. The property name has to contain a "$1", which gets replaced by whatever was matched by the "*".

### DIFF
--- a/pkg/utils/fieldpath/fieldpath.go
+++ b/pkg/utils/fieldpath/fieldpath.go
@@ -20,66 +20,103 @@ import (
 	"fmt"
 	"strings"
 
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
-// FormatMap formats map[string]string to a string.
-func FormatMap(m map[string]string) (fmtStr string) {
-	// output with keys in sorted order to provide stable output
-	keys := sets.NewString()
-	for key := range m {
-		keys.Insert(key)
-	}
-	for _, key := range keys.List() {
-		fmtStr += fmt.Sprintf("%v=%q\n", key, m[key])
-	}
-	fmtStr = strings.TrimSuffix(fmtStr, "\n")
-
-	return
-}
-
-// ExtractFieldPathAsString extracts the field from the given object
-// and returns it as a string.  The object must be a pointer to an
-// API type.
-func ExtractFieldPathAsString(obj interface{}, fieldPath string) (string, error) {
+// ExtractFieldPath extracts the field(s) from the given object
+// and returns the value as string, along with the expanded value of any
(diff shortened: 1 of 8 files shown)


pkg/utils/fieldpath/fieldpath.go
```diff
@@ -20,66 +20,103 @@ import (
...
 	accessor, err := meta.Accessor(obj)
 	if err != nil {
-		return "", err
+		return nil, nil, err
 	}
 
 	if path, subscript, ok := SplitMaybeSubscriptedPath(fieldPath); ok {
 		switch path {
 		case "metadata.annotations":
+			if strings.HasSuffix(subscript, "*") {
```
JoelColledge, line 43: The docs say:

> with a reference path containing a `*` character

This appears to implement "ending with a `*` character".

Which is right?

pkg/utils/fieldpath/fieldpath_test.go
```diff
@@ -128,7 +133,78 @@ func TestExtractFieldPathAsString(t *testing.T) {
...
+						"example.com/foo": "",
+						"example.com/baz": "bazbar",
+					},
+				},
+			},
+			expectedKeys:   []string{"baz", "foo"},
+			expectedValues: []string{"bazbar", ""},
+		},
+		{
+			name:      "ok - multivalue - empty value",
```
JoelColledge: Same test twice?

pkg/utils/fieldpath/fieldpath_test.go
```diff
@@ -180,8 +256,10 @@ func TestExtractFieldPathAsString(t *testing.T) {
 			}
 		} else if tc.expectedMessageFragment != "" {
 			t.Errorf("%v: expected error: %v", tc.name, tc.expectedMessageFragment)
-		} else if e := tc.expectedValue; e != "" && e != actual {
-			t.Errorf("%v: unexpected result; got %q, expected %q", tc.name, actual, e)
+		} else if (matched == nil) == (tc.expectedKeys == nil) && !slices.Equal(matched, tc.expectedKeys) {
```
JoelColledge: When `tc.expectedKeys != nil`, don't we want to validate that `matched != nil`?

pkg/utils/fieldpath/fieldpath.go
WanzenBug: Updated the docs, as ending on a `*` is actually the intention.

pkg/utils/fieldpath/fieldpath_test.go
WanzenBug: Updated for clarity, looks like I got lost in the weeds there. The intention is to test the distinction between returning `nil` and `[]` for deciding if this was a plain key or "glob".

JoelColledge (approved):

PR #702, opened 2024-08-28
Join properties

WanzenBug:
This replaces and expands the recent work on templating property resolution.
    
Using "expandFrom" users can both copy a collection of property based on a wildcard pattern, as well as create a single property based on all values of a wildcard pattern.
    
This reverts the changes to the "valuesFrom" property, as it was deemed too confusing. Instead, this behaviour is now possible with expandFrom.


docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -58,13 +58,22 @@ spec:
...
+
+Using `expandFrom` allows for using field references matching more than one field. Either specify a field that is
+already a map (`metadata.labels` or `metadata.annotations`), or select a subset by using `*` at the end of a key. Using
+`*` will select the keys and values matching the prefix up to the `*` character. There are two ways to use `expandFrom`:
+
+* Setting a `nameTemplate` will create one property per matched field. The property name is generated by taking
+  the `name` field and appending the expanded `nameTemplate`. `nameTemplate` supports the following expansions:
+  * `$1` is replaced with the field key, i.e. the part matched by the `+` character.
+  * `$2` is replaced with the value of the matched field.
+  The `valueTemplate` is expanded using the same replacements, and sets the property value.
```
JoelColledge, line 74: This line is formatted as part of the ``* `$2` ...`` bullet point.

docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -78,8 +87,11 @@ This examples sets three Properties on every satellite:
 * `AutoplaceTarget` (if set to `no`, will exclude the node from LINSTOR's Autoplacer) takes the value from the
   `piraeus.io/autoplace` annotation of the Kubernetes Node. If a node has no `piraeus.io/autoplace` annotation, the
   property will not be set.
-* `Aux/role/$1` copies all "node-role.kubernetes.io/*" label keys and values. For example, a worker node with the
-  `node-role.kubernetes.io/worker: "true"` label will have the `Aux/role/worker` set to `"true"`.
+* `Aux/role/` copies all "node-role.kubernetes.io/*" label keys and values. For example, a worker node with the
```
JoelColledge: "This examples sets three Properties on every satellite:" is no longer true.

api/v1/properties.go
```diff
@@ -47,26 +52,75 @@ type LinstorNodePropertyValueFrom struct {
 	NodeFieldRef string `json:"nodeFieldRef,omitempty"`
 }
 
+type LinstorNodePropertyExpandFrom struct {
+	LinstorNodePropertyValueFrom `json:",inline"`
+
+	// NameTemplate defines how the property key is expanded.
+	// If set, the templates is appended to the defined property name, creating multiple properties instead of one
```
JoelColledge:
```suggestion
	// If set, the template is appended to the defined property name, creating multiple properties instead of one
```

api/v1/properties.go
```diff
@@ -47,26 +52,75 @@ type LinstorNodePropertyValueFrom struct {
...
+			if err != nil {
+				result = append(result, field.Invalid(path.Child(strconv.Itoa(i), "expandFrom", "nodeFieldRef"), p.ExpandFrom.NodeFieldRef, fmt.Sprintf("Invalid reference format: %s", err)))
+			}
+
+			if keys == nil {
+				result = append(result, field.Invalid(path.Child(strconv.Itoa(i), "expandFrom", "nodeFieldRef"), p.ExpandFrom.NodeFieldRef, "Wildcard property required"))
+			}
+
+			if p.ExpandFrom.NameTemplate != "" && p.ExpandFrom.Delimiter != "" {
+				result = append(result, field.Invalid(path.Child(strconv.Itoa(i), "expandFrom"), p.ExpandFrom, "Expected exactly one of 'nameTemplate' and 'delimiter' to be set"))
```
JoelColledge: This only complains if both are set, rather than "exactly one" as the error message says. Do we need a check for neither being set? Or that allowed and we use a default delimiter?

docs/reference/linstorsatelliteconfiguration.md
JoelColledge: Strictly we should probably format all these example properties as `node-role.kubernetes.io/*` instead of "node-role.kubernetes.io/*". Might end up being too much code style, though :smile:

api/v1/properties.go
WanzenBug: Adapted the error message. I don't know what use an empty `nameTemplate` and empty `delimiter` would be, but an empty delimiter may be desired in some cases, and I don't want to switch to using `*string` for "empty" values.

JoelColledge (approved):

PR #722, opened 2024-10-21
Select IP families for LINSTOR Satellite connections

WanzenBug:
Some users reported running in a Cluster where Kubernetes assigns IPv6 addresses to the Satellite Pods, but those addresses are then not routable from the Controller Pod.
    
This leads to useless connection definitions, and in some cases to LINSTOR getting stuck trying to connect on the wrong network. While this is a misconfiguration on the Kubernetes side, as well as a bug on the LINSTOR side, we can still help these users by providing a way to filter the network interfaces we configure on the LINSTOR Satellites.
    
The option to set the "ipFamilies" on the Satellite Configuration is inspired by the option of the same name on the Kubernetes Service resources. When set, we filter the available addresses to only the specified IP family (IPv4 or IPv6). An empty list is the same as not setting any restrictions, keeping the existing behaviour.

In addition, ensure that interfaces once configured by the operator are also cleaned when they are no longer needed,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Option to select IP Family to use for LINSTOR control traffic.
 
+### Changed
+
+- Reconciliation of a Satellites network interfaces now also deletes unneeded interfaces.
+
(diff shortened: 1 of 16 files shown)


JoelColledge (approved):
Looks good.

PR #726, opened 2024-11-04
docs,how-to: restoring LINSTOR database backups

WanzenBug:
We automatically create backups, we should also inform people on how to use them.

### DIFF
--- a/docs/how-to/restore-linstor-db.md
+++ b/docs/how-to/restore-linstor-db.md
@@ -1,0 +1,91 @@
+# How to Restore a LINSTOR Database Backup
+
+This guide shows you how to restore a LINSTOR® Controller from a database backup. A backup is created automatically
+on every database migration of the default `k8s` database.
+
+> [!CAUTION]
+> Restoring from a backup means all changes made to the Cluster state after the backup was taken are lost.
+> This includes information about Persistent Volumes and Volume Snapshots that where created after the backup.
+
+To complete this guide, you should be familiar with:
+
+* using the `kubectl` command line tool to access the Kubernetes cluster.
+
+## Find the Latest Backup
+
+The backup is stored in Kubernetes Secrets. List all backups by using the following command:
+
+```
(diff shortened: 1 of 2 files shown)


docs/how-to/restore-linstor-db.md
```diff
@@ -0,0 +1,87 @@
...
+Select the backup you want to restore, making note of the name. For example, to restore the LINSTOR 1.27.0 version,
+we set:
+
+```
+$ BACKUP_NAME=linstor-backup-for-linstor-controller-db7fbfd95-zsfmm
+```
+
+## Temporarily stop the LINSTOR Controller Deployment
+
+To perform the upgrade, ensure that the LINSTOR Controller is shut down by scaling the Piraeus Operator and
```
JoelColledge: upgrade?

docs/how-to/restore-linstor-db.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Restore the Database
+
+Copy and unpack the selected backup to a local directory by using the following commands:
+
+```
+$ mkdir restore
+$ cd restore
+# Replace this with your selected backup name
```
JoelColledge: It is not that clear what "this" refers to in the long command.

docs/how-to/restore-linstor-db.md
```diff
@@ -0,0 +1,87 @@
+# How to Restore a LINSTOR Database Backup
+
+This guide shows you how to restore a LINSTOR® Controller from a database backup.
```
JoelColledge: Mention that this only applies to the default custom resources database type?

JoelColledge: It ought to be obvious, but warn that changes made since the backup was taken are lost? I guess most things will be re-created by the CSI driver. How reliably are we able to pick up the existing volumes with their data? Should normally work, I suppose.

JoelColledge (approved):

PR #751, opened 2024-12-20
Add an external controller metrics service

vladsf:
If a LINSTOR controller is external (piraeus chart installed with .Values.externalController.url set) then the external controller requires an Endpoint and an extra Service to make metrics available via ServiceMonitor.

This PR addresses the issue.

### DIFF
--- a/charts/piraeus/templates/external-metrics-service.yaml
+++ b/charts/piraeus/templates/external-metrics-service.yaml
@@ -30,8 +30,6 @@ spec:
   type: ExternalName
   externalName: {{ $ipAddr }}
   clusterIP: ""
-  selector:
-  {{- include "piraeus-operator.selectorLabels" . | nindent 4 }}
   ports:
     - name: metrics
       port: {{ $port }}
--- a/charts/piraeus/templates/external-metrics-service.yaml
+++ b/charts/piraeus/templates/external-metrics-service.yaml
@@ -1,23 +1,32 @@
 {{ if (.Values.externalController).url }}
 ---
 {{- $url := urlParse .Values.externalController.url -}}
+{{- $proto := regexSplit ":" (get $url "scheme") -1 -}}
 {{- $host := regexSplit ":" (get $url "host") -1 -}}
-{{- $ipAddr := index $host 0 -}}
+{{- $addr := index $host 0 -}}
 {{- $port := index $host 1 }}
+{{- if (regexMatch "^((25[0-5]|(2[0-4]|1\\d|[1-9]|)\\d)\\.?\\b){4}$" $host) -}}
+{{-   $addrType := "IPv4" -}}
+{{- else -}}
+{{-   $addrType := "FQDN" -}}
+{{- end -}}
 apiVersion: v1
-kind: Endpoints
+kind: EndpointSlice
 metadata:
   name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
   labels:
   {{- include "piraeus-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: linstor-external-controller
-subsets:
+    kubernetes.io/service-name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
+addressType: {{ $addrType }}
+ports:
+  - name: metrics
+    port: {{ $port }}
+    appProtocol: {{ $proto }}
+    protocol: TCP
+endpoints:
   - addresses:
-    - ip: {{ $ipAddr }}
-    ports:
-    - name: metrics
-      port: {{ $port }}
-      protocol: TCP
+    - {{ $addr }}
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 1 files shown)


charts/piraeus/templates/external-metrics-service.yaml
```diff
@@ -0,0 +1,39 @@
...
+  name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
+  labels:
+  {{- include "piraeus-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: linstor-external-controller
+spec:
+  type: ExternalName
+  externalName: {{ $ipAddr }}
+  clusterIP: ""
+  selector:
+  {{- include "piraeus-operator.selectorLabels" . | nindent 4 }}
```
WanzenBug: If you use a selector, Kubernetes will automatically try to manage the `Endpoint` and `EndpointSlice` resources. Which is probably not what you want: we manage the `Endpoint` already.

charts/piraeus/templates/external-metrics-service.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: Service
+metadata:
+  name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
+  labels:
+  {{- include "piraeus-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: linstor-external-controller
+spec:
+  type: ExternalName
+  externalName: {{ $ipAddr }}
+  clusterIP: ""
```
WanzenBug: According to https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors, you don't need to set `type`, `externalName` or `clusterIp`.

charts/piraeus/templates/external-metrics-service.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - ip: {{ $ipAddr }}
+    ports:
+    - name: metrics
+      port: {{ $port }}
+      protocol: TCP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
```
WanzenBug, line 34: This name also needs to be set on the `ServiceMonitor` resource instead of the usual controller metrics.

charts/piraeus/templates/external-metrics-service.yaml
```diff
@@ -0,0 +1,39 @@
...
+apiVersion: v1
+kind: Endpoints
+metadata:
+  name: {{ include "piraeus-operator.fullname" . }}-external-controller-metrics-service
+  labels:
+  {{- include "piraeus-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: linstor-external-controller
+subsets:
+  - addresses:
+    - ip: {{ $ipAddr }}
```
WanzenBug: This could also be a DNS name. We should use `EndpointSlice` instead, which has a `addressType` which we could set. I guess we would somehow automatically decide if the hostname is actually an IP address or not :thinking:

WanzenBug (changes requested):
Thanks! I left a few comments

charts/piraeus/templates/external-metrics-service.yaml
vladsf: Fixed.

charts/piraeus/templates/external-metrics-service.yaml
vladsf: Added support for FQDN

charts/piraeus/templates/external-metrics-service.yaml
vladsf: Fixed

charts/piraeus/templates/external-metrics-service.yaml
vladsf: This is not really clear to me. My patch has no ServiceMonitor. It is installed from config/extras.

vladsf: Should I simply remove "_external_" from *external-controller* names?

WanzenBug: Ah, sorry, I was mis-remembering the actual implementation. You can keep it as-is.

WanzenBug (changes requested):
I'm a bit slow today: actually, this whole thing should be moved to https://github.com/piraeusdatastore/helm-charts/tree/main/charts/linstor-cluster

There, we already have the `externalController` logic and even deploy a [monitoring](https://github.com/piraeusdatastore/helm-charts/blob/main/charts/linstor-cluster/templates/monitoring.yaml).

Sorry for the confusion. This helm chart also needs to be somehow moved there eventually.

vladsf:
> I'm a bit slow today: actually, this whole thing should be moved to https://github.com/piraeusdatastore/helm-charts/tree/main/charts/linstor-cluster
> 
> There, we already have the `externalController` logic and even deploy a [monitoring](https://github.com/piraeusdatastore/helm-charts/blob/main/charts/linstor-cluster/templates/monitoring.yaml).
> 
> Sorry for the confusion. This helm chart also needs to be somehow moved there eventually.

I have not found externalController logic there. I guess I should abandon this PR?

WanzenBug:
> I have not found externalController logic there. I guess I should abandon this PR?

Yes. Move it over to the other repo. There, users can already set `linstorCluster.externalController.url`. The chart here is just for deploying the operator, but stuff like setting up the `LinstorCluster` resource and setting up monitoring is controlled over there.

PR #765, opened 2025-01-22
automatically try to guess tag for satellites

WanzenBug:
When using an external LINSTOR Controller, the Satellite version might not match the default version we configure with the Operator. Make the Operator smart enough to handle the "happy path" of the users having a properly released LINSTOR Controller version.

To do this, we try to modify the first image configuration that tries to set the linstor-satellite image tag. This should always be the one we shipped with the Operator. Any config applied by the user would still take precedence.

### DIFF
--- a/internal/controller/linstorsatellite_controller.go
+++ b/internal/controller/linstorsatellite_controller.go
@@ -259,6 +259,24 @@ func (r *LinstorSatelliteReconciler) kustomizeNodeResources(ctx context.Context,
 		return nil, err
 	}
 
+	if lsatellite.Spec.ClusterRef.ExternalController != nil {
+		lc, err := linstorhelper.NewClientForCluster(
+			ctx,
+			r.Client,
+			r.Namespace,
+			&lsatellite.Spec.ClusterRef,
+			r.LinstorClientOpts...,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		err = imageversions.SetFromExternalCluster(ctx, lc.Client, cfg)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	imgs, precompiled := cfg.GetVersions(lsatellite.Spec.Repository, node.Status.NodeInfo.OSImage)
 
 	if precompiled {
@@ -294,14 +312,8 @@ func (r *LinstorSatelliteReconciler) reconcileLinstorSatelliteState(ctx context.
 		ctx,
 		r.Client,
 		r.Namespace,
-		lsatellite.Spec.ClusterRef.Name,
-		lsatellite.Spec.ClusterRef.ClientSecretName,
-		lsatellite.Spec.ClusterRef.CAReference,
-		lsatellite.Spec.ClusterRef.ExternalController,
-		append(
-			slices.Clone(r.LinstorClientOpts),
-			linstorhelper.Logr(log.FromContext(ctx)),
-		)...,
+		&lsatellite.Spec.ClusterRef,
+		r.LinstorClientOpts...,
 	)
 	if err != nil || lc == nil {
 		conds.AddError(conditions.Available, err)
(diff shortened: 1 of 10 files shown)


pkg/imageversions/dynamic.go
```diff
@@ -19,19 +21,35 @@ func FromConfigMap(ctx context.Context, client client.Client, name types.Namespa
...
+// Returns an error is the LINSTOR Controller could not be reached, or no LINSTOR Satellite image was found.
+func SetFromExternalCluster(ctx context.Context, client lclient.Client, configs Configs) error {
+	version, err := client.Controller.GetVersion(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch controller version: %w", err)
+	}
+
+	for _, config := range configs {
+		if v, ok := config.Components["linstor-satellite"]; ok {
+			v.Tag = "v" + version.Version
```
JoelColledge: Log that we are overriding this based on the Controller version?

pkg/imageversions/dynamic.go
```diff
@@ -19,19 +21,35 @@ func FromConfigMap(ctx context.Context, client client.Client, name types.Namespa
...
 	}
 
-	sort.Slice(cfgs, func(i, j int) bool {
-		return cfgs[i].source < cfgs[j].source
-	})
-
 	return cfgs, nil
 }
+
+// SetFromExternalCluster sets the LINSTOR Satellite image tag to the version of the LINSTOR client.
```
JoelColledge:
```suggestion
// SetFromExternalCluster sets the LINSTOR Satellite image tag to the version of the LINSTOR Controller.
```

internal/controller/linstorsatellite_controller.go
```diff
@@ -259,6 +259,30 @@ func (r *LinstorSatelliteReconciler) kustomizeNodeResources(ctx context.Context,
 		return nil, err
 	}
 
+	if lsatellite.Spec.ClusterRef.ExternalController != nil {
+		lc, err := linstorhelper.NewClientForCluster(
```
JoelColledge, line 263: Create a helper for the helper? There are several call sites that could be reduced to `LinstorSatelliteReconciler.NewClientForCluster(ctx, lsatellite)`.

The `r *LinstorNodeConnectionReconciler` case is probably too complex to unify with the `LinstorSatelliteReconciler` ones.

JoelColledge (commented):
Hooray for the tests :tada:

internal/controller/linstorsatellite_controller.go
WanzenBug: Did a half-measure, as I usually dislike having helpers that just call another function "in a special way". Downside is that we now need to construct a "ClusterReference" in the LinstorCluster case, but that seemed reasonable to me.

JoelColledge: I like it :+1:

pkg/imageversions/dynamic.go
```diff
@@ -19,19 +22,39 @@ func FromConfigMap(ctx context.Context, client client.Client, name types.Namespa
...
+		return fmt.Errorf("failed to fetch controller version: %w", err)
+	}
+
+	newTag := "v" + version.Version
+	logger := log.FromContext(ctx)
+
+	for name, config := range configs {
+		if v, ok := config.Components["linstor-satellite"]; ok {
+			logger.WithValues("configName", name, "oldTag", v.Tag, "newTag", newTag).Info("updating image version based on external cluster")
+			v.Tag = "v" + version.Version
```
JoelColledge: Nit:

```suggestion
			v.Tag = newTag
```

JoelColledge (dismissed):

JoelColledge (approved):

PR #767, opened 2025-01-27
kubebuilder: new webhook layout

WanzenBug:
New kubebuilder layout has the webhooks in a separate directory. This means a lot of moving code around with no clear benefit :-/


JoelColledge (approved):
I skimmed through the changes to the files that have been moved and changed offline. All looks fine :+1:

PR #779, opened 2025-02-25
tests: fix race condition in parallel tests

WanzenBug:
The CreateOrUpdateNode() call may modify the passed Node struct. This may lead to a concurrent map access, as the same Node is used in multiple tests.

Fix this by replacing the static sample node with a function that returns the sample node.


WanzenBug:
See https://github.com/piraeusdatastore/piraeus-operator/actions/runs/13516059559/job/37764861643 for such a test failure.

JoelColledge (approved):

PR #818, opened 2025-05-08
Impove satellite tests for interaction with LINSTOR Api

WanzenBug:
* Provide a fake LINSTOR API that can be used to test certain aspects of the satellite reconciler.
* Update some test methods to provide better error messages
* Add some additional configuration so we no longer need to explicitly set timeouts

### DIFF
--- a/internal/controller/linstorsatellite_test.go
+++ b/internal/controller/linstorsatellite_test.go
@@ -75,7 +75,7 @@ var _ = Describe("LinstorSatelliteReconciler", func() {
 				err = k8sClient.List(ctx, &satellites)
 				Expect(err).NotTo(HaveOccurred())
 				return satellites.Items
-			}, DefaultTimeout, DefaultCheckInterval).Should(BeEmpty())
+			}).Should(BeEmpty())
 
 			err = k8sClient.DeleteAllOf(ctx, &corev1.Node{})
 			Expect(err).NotTo(HaveOccurred())
@@ -96,7 +96,7 @@ var _ = Describe("LinstorSatelliteReconciler", func() {
 					return false
 				}
 				return condition.Status == metav1.ConditionTrue
-			}, DefaultTimeout, DefaultCheckInterval).Should(BeTrue())
+			}).Should(BeTrue())
 
 			Expect(satellite.Finalizers).To(ContainElement(vars.SatelliteFinalizer))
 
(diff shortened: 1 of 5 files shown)


internal/controller/linstorsatellite_test.go
```diff
@@ -186,48 +209,175 @@ var _ = Describe("LinstorSatelliteReconciler", func() {
...
+						Namespace:       Namespace,
+						Labels:          ds.Spec.Template.Labels,
+						Annotations:     ds.Spec.Template.Annotations,
+						OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(ds, schema.FromAPIVersionAndKind("apps/v1", "DaemonSet"))},
+					},
+					Spec: ds.Spec.Template.Spec,
+				}
+
+				pod.Spec.NodeName = ExampleNodeName
+				err := k8sClient.Create(ctx, pod)
```
JoelColledge, line 242: Why does the test now have to create a satellite pod?

pkg/fakelinstor/fakelinstor.go
```diff
@@ -0,0 +1,234 @@
...
+
+	if !slices.ContainsFunc(f.resources, func(exist lapi.ResourceWithVolumes) bool {
+		return exist.Name == rd && exist.NodeName == node
+	}) {
+		f.resources = append(f.resources, lapi.ResourceWithVolumes{
+			Resource: rc.Resource,
+		})
+	}
+
+	return nil, nil
```
JoelColledge, line 186: I guess it is deliberate and consistent with the real Controller to not return an error when the resource already exists?

JoelColledge (commented):
Looks like a sensible change to me. Just some questions for my understanding.

internal/controller/linstorsatellite_test.go
WanzenBug: We didn't need to create it up to now, because we did not have a controller API to begin with. Now that we have a "Controller", the operator needs the Pod to figure out the IP address of the satellite.

pkg/fakelinstor/fakelinstor.go
WanzenBug: Yes, a resource create on an existing node is a no-op in LINSTOR.

JoelColledge (approved):

PR #819, opened 2025-05-11
Fix dashboard for scrape duration

cometship:
The scrape duration shows no data currently. This PR fixes this by using a predictable job name.

### DIFF
--- a/config/extras/monitoring/linstor-satellite-monitor.yaml
+++ b/config/extras/monitoring/linstor-satellite-monitor.yaml
@@ -14,6 +14,10 @@ spec:
           sourceLabels:
             - __meta_kubernetes_pod_node_name
           targetLabel: node
+        - action: replace
+          sourceLabels:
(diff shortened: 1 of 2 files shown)


WanzenBug:
Thanks for the fix! Please ensure your commits are signed off (see [here](https://github.com/piraeusdatastore/piraeus-operator/pull/819/checks?check_run_id=42016663072))

Also, is the rewrite instructions necessary? I was under the impression that the `PodMonitor` name is used as the job name? So something like `job=~".*/linstor-satellite"` would do the trick.

cometship:
> Thanks for the fix! Please ensure your commits are signed off (see [here](https://github.com/piraeusdatastore/piraeus-operator/pull/819/checks?check_run_id=42016663072))

Can i use a no-reply email for the sign-off?

> So something like job=~".*/linstor-satellite" would do the trick.

Ahh true, that is more elegant i guess.

PR #822, opened 2025-05-14
Add `deletionPolicy` to Satellites

### LINKED ISSUE #519: [v2]: How to deal with node deletion/eviction
Currently, the implicit behaviour of the Operator is:

If a LinstorSatellite resource should no longer exist, because it's no longer matching the node labels, etc... delete it.
If a LinstorSatellite resource is deleted, it will be "finalized" by triggering node evacuation and either waiting for the node to be offline and then "lost", or if the node remains online: for all resources to be moved to other nodes.

This has already caused some pain to some users that were not expecting this behaviour. In particular, it makes it hard to "undelete" a satellite should that be desired.

WanzenBug:
Adds a configurable policy for deletion of satellites. We had some reports that the currently implemented "eviction" was unexpected, in particular because no automatic restore happened once the satellite was "undeleted". 

So now the default is to leave the satellite registered in LINSTOR, only cleaning up the kubernetes resources. The old default is still available as the `Evacuate` strategy.

Needs #818.

Closes #519 and #297.

### DIFF
--- a/api/v1/deletion_policy.go
+++ b/api/v1/deletion_policy.go
@@ -1,0 +1,23 @@
+package v1
+
+// DeletionPolicy configures the way LinstorSatellite resources are deleted.
+//
+// A LinstorSatellite may be deleted because:
+// * It no longer matches the affinity and node selector of the LinstorCluster resource.
+// * The node it references has been removed from Kubernetes.
+// * It was manually deleted outside the Operator.
+//
+// A LinstorSatellite may store the last copy of a volume, in which case it is not desirable to unconditionally remove
+// the satellite from the cluster. For this reason, the following deletion policies exist:
+//
+// * DeletionPolicyEvacuate will start evacuation of the LINSTOR Satellite and wait until it completes before removing the LinstorSatellite object, comparable to the "linstor node evacuate" command.
+// * DeletionPolicyRetain will retain the LINSTOR Satellite, keeping it registered in LINSTOR, but removing associated Kubernetes resources.
+// * DeletionPolicyDelete will remove the LINSTOR Satellite from the LINSTOR Cluster without prior eviction, comparable to the "linstor node lost" command.
+// +kubebuilder:validation:Enum:=Evacuate;Retain;Delete
+type DeletionPolicy string
+
+const (
+	DeletionPolicyEvacuate DeletionPolicy = "Evacuate"
+	DeletionPolicyRetain   DeletionPolicy = "Retain"
+	DeletionPolicyDelete   DeletionPolicy = "Delete"
+)
--- a/api/v1/linstorsatellite_types.go
+++ b/api/v1/linstorsatellite_types.go
@@ -62,6 +62,10 @@ type LinstorSatelliteSpec struct {
 	// If not set, the Operator will configure all families found in the Satellites Pods' Status.
 	// +kubebuilder:validation:Optional
 	IPFamilies []IPFamily `json:"ipFamilies,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Retain
+	DeletionPolicy DeletionPolicy `json:"deletionPolicy,omitempty"`
 }
 
 // LinstorSatelliteStatus defines the observed state of LinstorSatellite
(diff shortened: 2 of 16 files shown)


api/v1/deletion_policy.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+// DeletionPolicy specifies the way LINSTOR Satellite deletions are handled.
+// +kubebuilder:validation:Enum:=Evacuate;Retain;Delete
+type DeletionPolicy string
+
+const (
+	// DeletionPolicyEvacuate will start evacuation of the LINSTOR Satellite and wait until it completes to remove the LinstorSatellite object, comparable to the "linstor node evacuate" command.
```
JoelColledge: Perhaps put the documentation of the values in the doc for `DeletionPolicy` so that it appears in the CRD descriptions?

api/v1/deletion_policy.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+// DeletionPolicy specifies the way LINSTOR Satellite deletions are handled.
```
JoelColledge: "LINSTOR Satellite deletions" seems a bit ambiguous to me. Does it mean when the node is removed from Kubernetes? When the node is deleted from LINSTOR? When the LinstorSatellite custom resource is deleted? When the LinstorSatelliteConfiguration is changed such that a node is no longer included? When the Pod or DaemonSet is removed behind our back?

JoelColledge (commented):
Functionality looks good :+1:

api/v1/deletion_policy.go
WanzenBug: Great idea, thanks!

Now it looks like this:
```
$ kubectl explain linstorsatelliteconfigurations.spec.deletionPolicy
GROUP:      piraeus.io
KIND:       LinstorSatelliteConfiguration
VERSION:    v1

FIELD: deletionPolicy <string>
ENUM:
    Evacuate
    Retain
    Delete

DESCRIPTION:
    DeletionPolicy configures the way LinstorSatellite resources are deleted.
    
    A LinstorSatellite may be deleted because:
    * It no longer matches the affinity and node selector of the LinstorCluster
    resource.
    * The node it references has been removed from Kubernetes.
    * It was manually deleted outside the Operator.
    
    A LinstorSatellite may store the last copy of volume, in which case it is
    not desirable to unconditionally remove
    the satellite from the cluster. For this reason, the following deletion
    policies exist:
    
    * DeletionPolicyEvacuate will start evacuation of the LINSTOR Satellite and
    wait until it completes to remove the LinstorSatellite object, comparable to
    the "linstor node evacuate" command.
    * DeletionPolicyRetain will retain the LINSTOR Satellite, keeping it
    registered in LINSTOR, but removing associated Kubernetes resources.
    * DeletionPolicyDelete will remove the LINSTOR Satellite from the LINSTOR
    Cluster without prior eviction, comparable to the "linstor node lost"
    command.
```

api/v1/deletion_policy.go
WanzenBug: I think I've cleared it up with a more verbose explaination :)

api/v1/deletion_policy.go
```diff
@@ -0,0 +1,23 @@
+package v1
+
+// DeletionPolicy configures the way LinstorSatellite resources are deleted.
+//
+// A LinstorSatellite may be deleted because:
+// * It no longer matches the affinity and node selector of the LinstorCluster resource.
+// * The node it references has been removed from Kubernetes.
+// * It was manually deleted outside the Operator.
+//
+// A LinstorSatellite may store the last copy of volume, in which case it is not desirable to unconditionally remove
```
JoelColledge:
```suggestion
// A LinstorSatellite may store the last copy of a volume, in which case it is not desirable to unconditionally remove
```

api/v1/deletion_policy.go
```diff
@@ -0,0 +1,23 @@
...
+// the satellite from the cluster. For this reason, the following deletion policies exist:
+//
+// * DeletionPolicyEvacuate will start evacuation of the LINSTOR Satellite and wait until it completes to remove the LinstorSatellite object, comparable to the "linstor node evacuate" command.
```
JoelColledge:
```suggestion
// * DeletionPolicyEvacuate will start evacuation of the LINSTOR Satellite and wait until it completes before removing the LinstorSatellite object, comparable to the "linstor node evacuate" command.
```

JoelColledge (commented):
Looks great. Just some little nits.

JoelColledge (approved):

PR #825, opened 2025-05-23
make operator taint-aware

WanzenBug:
make operator taint-aware
    
    Taints are a way for Kubernetes to indicate that nodes should not run a
    particular workload. A particularly common Taint is:
    
      node-role.kubernetes.io/control-plane:NoSchedule
    
    to prevent workloads from running on the control plane nodes. Up to know,
    we specifically ignored these kinds of taints, letting Satellites and
    other workloads run even on those nodes.
    
    This often causes confusion and issues, in particular on some cloud platforms,
    where these control plane nodes really cannot run any workload.
    
    So instead, we just behave like any other workload, using the default
    tolerations. Since we basically behave like a DaemonSet for Satellites,
    we copy the default tolerations for a DaemonSet.
    
    To get back the old behaviour, we add a new LinstorCluster.spec.tolerations
    setting, allowing users to add additional tolerations.

also some small commits for tests and CRDs.


api/v1/linstorcluster_types.go
```diff
@@ -44,6 +44,13 @@ type LinstorClusterSpec struct {
 	// +kubebuilder:validation:Optional
 	NodeAffinity *corev1.NodeSelector `json:"nodeAffinity,omitempty"`
 
+	// Tolerations selects the nodes on which LINSTOR Satellites will be deployed.
```
JoelColledge, line 47: I'm wondering if we might want this to be part of `LinstorSatelliteConfiguration` instead of `LinstorCluster`? Might we want different tolerations for different satellites?

internal/controller/linstorcluster_controller.go
```diff
@@ -332,6 +339,16 @@ func (r *LinstorClusterReconciler) kustomizeControllerResources(lcluster *piraeu
 		patches = append(patches, p...)
 	}
 
+	if len(lcluster.Spec.Tolerations) > 0 {
+		t := tolerations.MergeTolerations(tolerations.HAControllerTolerations, lcluster.Spec.Tolerations)
```
JoelColledge: It looks like the `HAControllerTolerations` are only applied when additional tolerations have been added. Shouldn't they always be applied?

internal/controller/linstorcluster_test.go
```diff
@@ -302,6 +315,92 @@ var _ = Describe("LinstorCluster controller", func() {
...
+					// Another "core" taint we ignore by default.
+					{Key: corev1.TaintNodeUnschedulable, Effect: corev1.TaintEffectNoSchedule},
+					// A Taint we manually tolerate later.
+					{Key: "example.com/manual-taint", Effect: corev1.TaintEffectPreferNoSchedule},
+					// A Taint we never toleration
+					{Key: "example.com/untolerated-taint", Effect: corev1.TaintEffectNoExecute},
+				}
+
+				for i := range nodes.Items {
+					// Apply two taints to the first nodes, three to the second, all four to the third
```
JoelColledge:
```suggestion
					// Apply two taints to the first node, three to the second, all four to the third
```

internal/controller/linstorcluster_test.go
```diff
@@ -302,6 +315,92 @@ var _ = Describe("LinstorCluster controller", func() {
...
+				Expect(nodes.Items).Should(HaveLen(3))
+
+				taintsToAdd := []corev1.Taint{
+					// A HA Controller taint we ignore by default.
+					{Key: metadata.NodeForceIoErrorTaint, Effect: corev1.TaintEffectNoSchedule},
+					// Another "core" taint we ignore by default.
+					{Key: corev1.TaintNodeUnschedulable, Effect: corev1.TaintEffectNoSchedule},
+					// A Taint we manually tolerate later.
+					{Key: "example.com/manual-taint", Effect: corev1.TaintEffectPreferNoSchedule},
+					// A Taint we never toleration
```
JoelColledge:
```suggestion
					// A Taint we never tolerate
```

api/v1/linstorcluster_types.go
WanzenBug: I guess one could want that. The trouble is that the `LinstorCluster` part is also the one that selects which nodes should run Satellites at all. The `LinstorClusterConfiguration` is just instructing how a particular satellite on a particular node should be run.

This is also why there is a node selector/affinity setting for `LinstorCluster`, where you basically set the "upper bound" on nodes that should run Piraeus. In this sense, the additional tolerations are just the completion of that feature.

This is actually something that often trips up our users, i.e. if they just create a `LinstorCluster`, it will already create the satellites, there is no need to also use a `LinstorSatelliteConfiguration`. I should probably get around to filling the "how does Piraeus work" part of the documention :sweat_smile:

internal/controller/linstorcluster_controller.go
WanzenBug: I guess so. It is actually a bit tricky, as I think we should apply the "user" tolerations for all workloads (LINSTOR Controller, CSI Controller), but the "daemon set" tolerations should only apply to HA Controller, CSI Nodes and Satellites. 

I had the code set up a bit differently before, but now that should not be hard to implement.

WanzenBug: Also, for the "no user tolerations" case, they are already defined in the .yaml. Removing that now, as it gets confusing.

pkg/utils/tolerations/tolerations.go
```diff
@@ -0,0 +1,116 @@
...
+		{
+			Key:      metadata.NodeLostQuorumTaint,
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		},
+	}
+
+	// FailOverTimeOut is the number of seconds the LINSTOR Controller should tolerate fail-over worthy Taints.
+	FailOverTimeOut = int64(30)
+	// LinstorControllerTolerations tolerate fail-over relevent taints only up to FailOverTimeOut
```
JoelColledge:
```suggestion
	// LinstorControllerTolerations tolerate fail-over relevant taints only up to FailOverTimeOut
```

internal/controller/linstorcluster_controller.go
```diff
@@ -453,6 +468,14 @@ func (r *LinstorClusterReconciler) kustomizeCSIControllerResources(lcluster *pir
 		patches = append(patches, p...)
 	}
 
+	t := tolerations.MergeTolerations(tolerations.LinstorControllerTolerations, tolerations.HAControllerTolerations, lcluster.Spec.Tolerations)
```
JoelColledge: Why does `linstor-csi-controller` get the `LinstorControllerTolerations`? The comment on `FailOverTimeOut` indicates that it applies only to the LINSTOR Controller.

JoelColledge (commented):
Good point with the excess tolerations in the yaml files.

JoelColledge (approved):

PR #833, opened 2025-06-10
Add image digests & `RELATED_IMAGE_*` env support

WanzenBug:
- Allow pinning images in the image configuration to a specific digest.
- Support overriding image configuration directly on the Operator Deployment using `RELATED_IMAGE_*` environment variables.

### DIFF
--- a/docs/explanation/image-configuration.md
+++ b/docs/explanation/image-configuration.md
@@ -1,0 +1,183 @@
+# How Piraeus Operator Selects Image Versions for Deployed Resources
+
+Piraeus Datastore consists of many different components, most of which follow their own release cycle. For this reason,
+Piraeus Operator comes with a default configuration that uses the latest versions of all components at the time of
+release.
+
+## The Default Image Configuration
+
+The default deployment of Piraeus Datastore creates a special ConfigMap resource named `piraeus-operator-image-config`.
+It contains two entries that represent YAML documents:
+
+* [`0_piraeus_datastore_images.yaml`](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/config/manager/0_piraeus_datastore_images.yaml),
+  which contains the image configuration for all components developed under the Piraeus Datastore umbrella.
+* [`0_sig_storage_images.yaml`](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/config/manager/0_sig_storage_images.yaml),
+  which contains the image configuration for all components developed under the [Kubernetes CSI](https://kubernetes-csi.github.io/) umbrella.
+
+The YAML documents follow the following scheme:
+
+```yaml
+base: quay.io/piraeusdatastore
+```
+
+The "base" repository. All images configured in this YAML document will start with `quay.io/piraeusdatastore/`.
+
+```yaml
+components:
+  linstor-controller:
+    tag: v1.31.0
+    image: piraeus-server
+```
+
+The `components` are the image names used in the workload resources. `image` is the image repository to use in the given
+"base" registry. `tag` is the image tag. There is also the option to set a `digest`, which takes precedence over tags,
+but is not used in the default configuration.
+
+These images are referenced in the Piraeus resources. For example, the LINSTOR Controller deployment has the
+[following container configuration](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/pkg/resources/cluster/controller/controller-deployment.yaml),
+which will be replaced given the above configuration with `quay.io/piraeusdatastore/piraeus-server:v1.31.0`:
+
+```yaml
+# ...
+      containers:
+        - name: linstor-controller
+          # will be replaced with:
+          # image: quay.io/piraeusdatastore/piraeus-server:v1.31.0
+          image: linstor-controller
+# ...
+```
+
+For cases where it is necessary to use different images based on the node the workload is running on, there is the
+option to set an additional `match` configuration:
+
+```yaml
+components:
+  drbd-module-loader:
+    tag: v9.2.13
+    image: drbd9-noble
+    match:
+      - osImage: Red Hat Enterprise Linux Server 7\.
+        image: drbd9-centos7
+      - osImage: Red Hat Enterprise Linux 8\.
+        image: drbd9-almalinux8
+# ...
+```
+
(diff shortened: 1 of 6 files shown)


pkg/imageversions/versions.go
```diff
@@ -55,36 +61,63 @@ func (f *Config) GetVersions(base string, osImage string) ([]kusttypes.Image, bo
...
+			}, matchRule.Precompiled
 		}
 	}
 
 	if img.Image == "" {
-		return "", "", false
+		return nil, false
+	}
+
+	if img := getOverrideForImageFromEnv(img.Image); img != nil {
```
JoelColledge: The commit message indicates that the convention for these `RELATED_IMAGE_` variables is to use the component name as the key. I can see that this does not work when overriding the image for a "match", but perhaps we should respect that convention for other images? E.g. `RELATED_IMAGE_drbd-module-loader` instead of `RELATED_IMAGE_drbd9-noble` to change the fallback.

To be very verbose, we could even use `RELATED_IMAGE_drbd-module-loader_drbd9-almalinux8` for overriding a match.

WanzenBug: :thinking: I like that idea. I was struggeling with how to properly support the case for "match", but using both component name and image name for this seems like a good way.

JoelColledge: Whatever we decide, I guess we should document it briefly somewhere. Even if we don't expect this to be a normal way for users to configure the operator.

WanzenBug: Yeah, working on that right now. Will update this PR once I have something written down.

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+release.
+
+## The Default Image Configuration
+
+The default deployment of Piraeus Datastore creates a special ConfigMap resource named `piraeus-operator-image-config`.
+It contains two entries that represent YAML documents:
+
+* [`0_piraeus_datastore_images.yaml`](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/config/manager/0_piraeus_datastore_images.yaml),
+  which contains the image configuration for all components developed under the Piraeus Datastore umbrella.
+  * [`0_sig_storage_images.yaml`](https://github.com/piraeusdatastore/piraeus-operator/tree/v2/config/manager/0_sig_storage_images.yaml),
```
JoelColledge: Indent?

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+```yaml
+components:
+  linstor-controller:
+    tag: v1.31.0
+    image: piraeus-server
+```
+
+The `components` are the image names used in the workload resources. `image` is the image repository to use in the given
+"base" registry. `tag` is the image tag. There is also the option to set a `digest`, which takes precedence over tags,
+but is not used in the default configuration. For example, the LINSTOR Controller deployment has the
```
JoelColledge:
```suggestion
but is not used in the default configuration.

These images are referenced in the Piraeus resources. For example, the LINSTOR Controller deployment has the
```

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+When a `match` section is present, the Operator will first try to match the Kubernetes Node "OS Image" (as reported by
+[`status.nodeInfo.osImage`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeStatus))
+against the given `osImage`. If a match is found, the matches' `image` is used instead of the component one. The
+`osImage` value is a [regular expression](https://pkg.go.dev/regexp). This is currently only used for the
+`drbd-module-loader`.
+
+## External Cluster Automations
+
+Apart from the matching of "OS Image" for the drbd-module-loader, there is a second automation built into the Operator:
+When using an external LINSTOR cluster, the Operator will match the LINSTOR version as reported by the external cluster
```
JoelColledge: Link to the corresponding how-to?

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+with matching tags.
+
+For example, assuming the external LINSTOR Cluster is running 1.31.2, as reported here:
+
+```
+$ linstor controller version
+linstor controller 1.31.2; GIT-hash: 6a9a4bb37f547ff73317585a3efa217523e01f43
+```
+
+The operator will automatically replace the default tags of the `linstor-controller` and `linstor-satellite` components
```
JoelColledge: Is the `linstor-controller` image used when an external controller is configured?

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+| csi-attacher       | registry.k8s.io/sig-storage/csi-attacher:v4.8.1 | registry.example.com/piraeus/csi-attacher:v4.8.1    |
+
+### Overriding the Image Configuration by Adding New YAML Documents
+
+The Operator will read all YAML documents in the Image Configuration ConfigMap. The files are read in ascending order,
+so providing a `1_override.yaml` document ensures the Operator will read this document after the default configuration.
+Since the Operator applies a "last configuration wins" algorithm for components, this document overrides the default
+configuration.
+
+The following example replaces the `linstor-controller` with version and 1.31.1 and pins the `linstor-satellite` image
```
JoelColledge:
```suggestion
The following example replaces the `linstor-controller` with version 1.31.1 and pins the `linstor-satellite` image
```

docs/explanation/image-configuration.md
```diff
@@ -0,0 +1,175 @@
...
+metadata:
+  name: piraeus-operator-image-config
+data:
+  0_piraeus_datastore_images.yaml: |
+    # ...
+  0_sig_storage_images.yaml: |
+    # ...
+  1_override.yaml: |
+    ---
+    base:
```
JoelColledge: Seems to have been cut off.

JoelColledge (commented):
Great docs!

The commit message of the second commit still references the old env var format.

docs/explanation/image-configuration.md
WanzenBug: You are absolutely right, it is not used :sweat_smile:

JoelColledge (approved):

PR #836, opened 2025-06-12
monitoring: exclude Inconsistent resources from drbdDeviceNotUpToDate…

WanzenBug:
… alert

This alert should report resources in "unexpected" states. However, the way we handle restoring from a snapshot, or even the addition of new peers to a resource might be triggering this alert by accident.

The issue is that when new peers are added to an existing volume, it will cause a partial or full resync, during which the volume stays Inconsistent. This is expected and fine. We already have a separate alert "drbdResourceResyncWithoutProgress" that reports disks that are Inconsistent but made no progress on the resync front.

### DIFF
--- a/config/extras/monitoring/alerts.yaml
+++ b/config/extras/monitoring/alerts.yaml
@@ -73,7 +73,7 @@ spec:
           annotations:
             description: |
               DRBD device "{{ $labels.name }}" on "{{ $labels.node }}" has unexpected device state "{{ $labels.drbd_device_state }}".
-          expr: drbd_device_state{drbd_device_state!~"UpToDate|Diskless"} > 0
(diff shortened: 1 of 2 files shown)


WanzenBug:
@JoelColledge out of interest, do you see any issues with the "drbdResourceResyncWithoutProgress" alert with the new DRBD 9.3 "resync without replication"? I guess in theory it could report an increase in the number of out-of-sync bytes with certain write patterns?

JoelColledge (approved):

JoelColledge:
> @JoelColledge out of interest, do you see any issues with the "drbdResourceResyncWithoutProgress" alert with the new DRBD 9.3 "resync without replication"? I guess in theory it could report an increase in the number of out-of-sync bytes with certain write patterns?

It could theoretically happen. I think it would only occur in a fairly pathological situation. Only real-world data will show really.

PR #846, opened 2025-07-07
satellites: keep running on "NoSchedule" nodes

WanzenBug:
We try to emulate DaemonSet logic for our Satellites, including the toleration of some default taints. However, our emulation was not perfect in two cases:

* DaemonSets only look at "NoExecute" and "NoSchedule" taints, ignoring "PreferNoSchedule" taints. This becomes noticable when certain distributions run upgrades (such as OpenShift or Cluster API managed k8s), as they generally set a "PreferNoSchedule" taint on nodes they consider outdated.
* When a Pod in a DaemonSet is already assigned to a node, it also does not get deleted when a "NoSchedule" taint is added. Instead, the Pod keeps running until manual deletion, or a NoExecute taint gets added.

This is now also implemented for our LinstorSatellite resources. A test is also added to verify the behaviour of keeping Pods running in the face of added "*NoSchedule" taints.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Image configuration for RHEL10 clones.
+- Do not remove existing satellites for `*NoSchedule` taints.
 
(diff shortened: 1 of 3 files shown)


JoelColledge (approved):
Looks good.

PR #851, opened 2025-07-16
Fix taint tests

WanzenBug:
* Ensure that we run our kubernetes API GC after every test
* Ensure that a deleted satellite on a `*NoSchedule` node does not get recreated.

### DIFF
--- a/internal/controller/linstorcluster_controller.go
+++ b/internal/controller/linstorcluster_controller.go
@@ -161,28 +161,13 @@ func (r *LinstorClusterReconciler) reconcileAppliedResource(ctx context.Context,
 		return err
 	}
 
-	satelliteTolerations := tolerations.MergeTolerations(tolerations.DefaultDaemonSetTolerations, tolerations.HAControllerTolerations, lcluster.Spec.Tolerations)
-	satelliteNodes.Items = slices.DeleteFunc(satelliteNodes.Items, func(node corev1.Node) bool {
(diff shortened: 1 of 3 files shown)


internal/controller/linstorcluster_controller.go
```diff
@@ -199,6 +199,18 @@ func (r *LinstorClusterReconciler) reconcileAppliedResource(ctx context.Context,
 			return err
 		}
 
+		// Special case for satellites on "NoSchedule" nodes. There is a possible race where we:
+		// * Performed a successful check that the satellite already exists
+		// * LinstorSatellite gets removed
+		// * We only now end up here
+		// We would now re-create the satellite, unless we ensure the UID is already set, in which case
+		// the Patch() will fail. We then retry, but this time we will see that the satellite has been removed.
+		if i := slices.IndexFunc(existingSatellites.Items, func(s piraeusiov1.LinstorSatellite) bool {
```
JoelColledge: We already check which satellite nodes have an existing satellite above. I feel like it would be cleaner to do this UID propagation there. However, I can't see how that would work. We would need to somehow attach the UID to the satellite Node and then embed it in the ResMap. What do you think?

WanzenBug: Done that now... I also moved the taint/tolerations logic a bit closer to where the generate the Satellites. Using a debugger, I verified that the UID is set as expected, and that we get the expected error on applying a deleted satellite.

JoelColledge (approved):
I find this cleaner :+1:

PR #852, opened 2025-07-16
add LINSTOR Affinity Controller as deployed resource

WanzenBug:
Like the HA Controller, the Affinity Controller is a generally useful piece in the Piraeus "ecosystem". It keeps the PV affinity in sync with LINSTOR resources, ensuring that during node evacuations, the PVs remain reachable from the perspective of Kubernetes.


WanzenBug:
Rather large diff, but it should be mostly static configuration....

JoelColledge (approved):
All looks good to me.

PR #854, opened 2025-07-17
satellites: wait until pods are ready during registration

WanzenBug:
This is a mainly cosmetic change, but it will cause the Satellite to only be registered when the Satellite process is ready to receive connections. This prevents the Satellites from showing up as OFFLINE during start up.


JoelColledge (approved):
Seems like a reasonable condition to wait for.

PR #855, opened 2025-07-18
satellite: graceful evacuation when using ClusterAPI

WanzenBug:
In a ClusterAPI-managed cluster, we want to support graceful evacuation of nodes. In particular, we want to ensure a smooth transfer of resources when new machines are rolled out and old ones are shut down.

To do that, we adapt the existing evacuation of satellites to better handle resource migration. The new evacuation performs 8 steps:

1. We ensure that existing PVs are reschedulable to other nodes. This is necessary for volumes using the "allowRemoteVolumeAccess: false" parameter. Otherwise, these PVs remain stuck to the node that is going to be removed.
2. We wait for all LINSTOR Satellite resources to report a ready state. This ensures that during a rolling restart/update of a cluster, the replacement node is already ready from LINSTOR side.
3. Allow ClusterAPI to drain the machine, moving workloads to replacement nodes.
4. Wait for all PVs that were previously attached on the evacuating Node to be attached on other Nodes. This helps LINSTOR with the following node evacuation.
5. Signal LINSTOR to start evacuating the Satellite. In particular, since the previously attached PVs are now used on a different node, LINSTOR will try to place the new replica on that particular node, if possible.
6. Wait for evacuationt to complete.
7. Remove any temporary annotations used to mark PVs are reschedulable. The "allowRemoteVolumeAccess: false" resources should have new diskfuls now, letting the Affinity Controller re-apply the strict node affinity.
8. Allow ClusterAPI to remove the node, also removing the Satellite.


WanzenBug:
@JoelColledge Sorry for the large diff, but I don't really see how this could be made in smaller increments. I'm not really happy with the duplication/general flow of the ClusterAPI/Machine stuff within the "reconcileSatelliteState" method, but I also don't see how it could be anywhere else. There is also the duplication with "deleteSatellite", but again, I don't have a clue on how to do better. 

I'm working on a explanatory text for how deletion policies and `deletionPolicy: Evacuate` work in particular.

cmd/main.go
```diff
@@ -104,9 +108,50 @@ func main() {
...
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		clientCfg, err := clientcmd.NewClientConfigFromBytes(rawCfg)
+		if err != nil {
+			setupLog.Error(err, "unable to parse cluster api kubeconfig")
+			os.Exit(1)
+		}
+
+		machineClientCfg, err = clientCfg.ClientConfig()
+		if err != nil {
+			setupLog.Error(err, "unable to parse cluster api kubeconfig")
```
JoelColledge: This error message is used twice. Make them distinct to help identify the cause of errors?

go.mod
```diff
@@ -52,9 +55,13 @@ require (
 	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/google/pprof v0.0.0-20250403155104-27863c87afa6 // indirect
 	github.com/google/uuid v1.6.0 // indirect
+	github.com/haySwim/data v0.2.0 // indirect
```
JoelColledge, line 58: I wonder how this indirect dependency got in? A module from Hayley, untouched for over 6 years :smile:

Just to be clear: I don't see any cause for concern with the dependency itself.

internal/controller/linstorsatellite_controller.go
```diff
@@ -512,9 +558,14 @@ func (r *LinstorSatelliteReconciler) reconcileStoragePools(ctx context.Context,
 	return nil
 }
 
-func (r *LinstorSatelliteReconciler) deleteSatellite(ctx context.Context, lsatellite *piraeusiov1.LinstorSatellite) error {
+func (r *LinstorSatelliteReconciler) deleteSatellite(ctx context.Context, lsatellite *piraeusiov1.LinstorSatellite, node *corev1.Node, conds conditions.Conditions) (string, bool, error) {
```
JoelColledge, line 572: Document the return values?

pkg/evacuation/evacuation.go
```diff
@@ -0,0 +1,364 @@
...
+				continue
+			}
+			return "", false, fmt.Errorf("failed to get PV %s: %w", pvName, err)
+		}
+
+		toUpdate := false
+
+		if pv.Annotations == nil {
+			pv.Annotations = make(map[string]string)
+			toUpdate = true
```
JoelColledge: Is converting a nil value into an empty map a reason to update? When we update with an empty map and then `Get` again, do we get an empty map back?

internal/controller/linstorsatellite_controller.go
```diff
@@ -311,6 +317,32 @@ func (r *LinstorSatelliteReconciler) kustomizeNodeResources(ctx context.Context,
 }
 
 func (r *LinstorSatelliteReconciler) reconcileLinstorSatelliteState(ctx context.Context, lsatellite *piraeusiov1.LinstorSatellite, node *corev1.Node, conds conditions.Conditions) error {
+	machine, err := r.MachineClient.GetMachineForNode(ctx, node)
```
JoelColledge, line 325: Perhaps a brief comment explaining why it is OK to call this regardless of whether ClusterAPI integration is configured?

pkg/evacuation/evacuation.go
```diff
@@ -0,0 +1,364 @@
...
+	if len(unattachedPVs) > 0 {
+		return fmt.Sprintf("Waiting for PVs to reattach on other nodes: %s", strings.Join(unattachedPVs, ", ")), false, nil
+	}
+
+	return "", true, nil
+}
+
+func evacuateNode(ctx context.Context, lclient *lapi.Client, node *lapi.Node) error {
+	// We retry evacuation every time. LINSTOR needs multiple kicks until it actually starts moving all resources in
+	// some cases.
```
JoelColledge, line 239: Perhaps this needs fixing in LINSTOR :joy:

pkg/evacuation/evacuation.go
```diff
@@ -0,0 +1,364 @@
...
+
+		attachedPVs = append(attachedPVs, *attachment.Spec.Source.PersistentVolumeName)
+	}
+
+	return attachedPVs
+}
+
+func getAttachedNodes(pvName string, attachments []storagev1.VolumeAttachment) []string {
+	var attachedNodes []string
+	for _, attachment := range attachments {
```
JoelColledge, line 322: It looks like we iterate over all the attachments in the cluster for all the PVs on the node. It seems realistic to me that we could have 100s of PVs on a node and 1000s of attachments in total. That makes something on the order of millions of iterations of the inner loop each time we reconcile while in this state. Right? Should we do it more efficiently? I'm sure it is possible with one iteration over the attachments, building a map by PV or something.

pkg/evacuation/evacuation_test.go
```diff
@@ -0,0 +1,434 @@
...
+		TestNodeA, TestNodeB, MachineA, MachineB,
+		createSatellite(TestNodeA, true),
+		createSatellite(TestNodeB, true),
+		localPV, createAttachment(TestNodeB, localPV),
+		unattachedLocalPV,
+		remotePV, createAttachment(TestNodeB, remotePV),
+		unattachedRemotePV,
+	)
+
+	// Create fake linstor resources, let the "unattached" ones be still on node A.
```
JoelColledge: Update copied comment.

JoelColledge (commented):
I have looked through this and it seems pretty clean to me :+1:

go.mod
WanzenBug: I think that is because we now use some additional packages from the `linstor-csi` project. We use that project there, I guess you could consider that project complete :smile:

pkg/evacuation/evacuation.go
WanzenBug: > I'm sure it is possible with one iteration over the attachments, building a map by PV or something.

Done just that :+1:

pkg/evacuation/evacuation.go
WanzenBug: It is not. It would generally cause an update, as with an empty map we are guaranteed to also trigger the following updates, but I removed this here still.

WanzenBug:
@JoelColledge please re-review

I added docs around the deletion process. I also fixed a super small issue in our webhooks that are not actually related to this feature, just noticed it during testing...

docs/explanation/deletion-policies.md
```diff
@@ -0,0 +1,117 @@
...
+Using the `Evacuate` deletion policy, Piraeus Datastore will first ensure the configured number of replicas for all
+volumes remains available before removing the Satellite. The Satellite is only removed after all local replicas and
+snapshots are removed.
+
+In order to keep potential fail-over times short, Piraeus Datastore performs the following steps during evacuation:
+
+1. Piraeus Datastore searches for PersistentVolumes (PVs) that:
+    * Are attached on the Node to be evacuated and marks them with `wait-for-reattach.evacuation.piraeus.io/<node-name>`
+      annotation. It will later use this annotation to wait for the PV to be reattached.
+    * In addition, if the PVS are also either using the `allowRemoteVolumeAcesss: "false"` policy or are otherwise bound
```
JoelColledge:
```suggestion
    * In addition, if the PVs are also either using the `allowRemoteVolumeAcesss: "false"` policy or are otherwise bound
```

docs/explanation/deletion-policies.md
```diff
@@ -0,0 +1,117 @@
...
+      to the Node to be evacuated in their `.spec.nodeAffinity`, they are annotated with
+      `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity Controller to temporarily mark the PV as
+      accessible from any Node.
+2. Piraeus Datastore will wait for any other `LinstorSatellite` resource to become available. This ensures that during
```
JoelColledge:
```suggestion
2. Piraeus Datastore will wait for all other `LinstorSatellite` resources to become available. This ensures that during
```

Right?

docs/explanation/deletion-policies.md
```diff
@@ -0,0 +1,117 @@
...
+   rolling upgrades of the infrastructure, it waits until a replacement node for the node to be evacuated is available.
+3. If using ClusterAPI, the `pre-drain.delete.hook.machine.cluster.x-k8s.io/linstor-prepare-for-drain` annotation on the
+   Machine is removed. This will cause ClusterAPI to start draining the node.
+4. Piraeus Datastore will wait for all PVs that were marked in Step 1 to be reattached on new nodes. This ensures that
+   LINSTOR can choose the optimal replica placement for the next step.
+5. Piraeus Datastore will signal LINSTOR to evacuate the node. LINSTOR will create replacement resources for all diskful
+   and diskless resources on the node. Volumes that are currently in use on a diskless will get a local replica, if that
+   is possible based on the available storage pools.
+6. Piraeus Datastore will wait for LINSTOR to complete the evacuation. The evacuated Node no longer has any resources
+   or snapshot.
```
JoelColledge:
```suggestion
   or snapshots.
```

docs/explanation/deletion-policies.md
```diff
@@ -0,0 +1,117 @@
...
+### `Delete` Policy
+
+When using the `Delete` policy, Piraeus Datastore will remove the Satellite from the LINSTOR Cluster in addition to
+removing the associated Kubernetes resources. The Satellite will no longer appear in the list of LINSTOR nodes and
+any resources on it are removed.
+
+!!! warning "There are no checks preventing deletion of the last replica of a volume"
+
+    Using this policy means that there are no checks the prevent the deletion of the last replica of a volume. While
+    LINSTOR automatically tries to maintain the requested number of replicas, does so only periodically.
```
JoelColledge:
```suggestion
    LINSTOR automatically tries to maintain the requested number of replicas, it does so only periodically.
```

pkg/evacuation/evacuation.go
```diff
@@ -0,0 +1,367 @@
...
+	return "", true, nil
+}
+
+// preparePVsforEvacuation is the first step of node evacuation.
+//
+// It checks for PVs attached to the current node and:
+// * marks them with annotations so later steps can find them.
+// * ensures that "local" PVs are temporarily reschedulable during evacuation.
+func preparePVsforEvacuation(ctx context.Context, cl client.Client, nodeName string) (string, bool, error) {
+	var volumeAttachementList storagev1.VolumeAttachmentList
```
JoelColledge:
```suggestion
	var volumeAttachmentList storagev1.VolumeAttachmentList
```

pkg/evacuation/evacuation.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func waitForReattach(ctx context.Context, cl client.Client, nodeName string) (string, bool, error) {
+	var pvs corev1.PersistentVolumeList
+	err := cl.List(ctx, &pvs)
+	if err != nil {
+		return "", false, err
+	}
+
+	var volumeAttachementList storagev1.VolumeAttachmentList
```
JoelColledge:
```suggestion
	var volumeAttachmentList storagev1.VolumeAttachmentList
```

JoelColledge (commented):
The new explanation doc is great :100:

Just noticed a few tiny details.

WanzenBug:
Thanks for the attentive review! I squashed the suggested changes into the right commits.

JoelColledge (approved):

PR #870, opened 2025-09-01
evacuation: add option to remove PVCs on evacuation

WanzenBug:
For some types of volumes, such as "scratch" volumes, it might be desirable to automatically delete the volume instead of migrating.

To allow that, scan the PV and as a fallback the RD and RG. There, we search for specific annotations/properties. This allows users to set either the property on the storage class, or update the PV when needed to trigger deletion.

The deletion of the volume will only complete after the currently using Pod released the volume.

### DIFF
--- a/docs/explanation/deletion-policies.md
+++ b/docs/explanation/deletion-policies.md
@@ -66,12 +66,22 @@ snapshots are removed.
 In order to keep potential fail-over times short, Piraeus Datastore performs the following steps during evacuation:
 
 1. Piraeus Datastore searches for PersistentVolumes (PVs) that:
-    * Are attached on the Node to be evacuated and marks them with `wait-for-reattach.evacuation.piraeus.io/<node-name>`
-      annotation. It will later use this annotation to wait for the PV to be reattached.
-    * In addition, if the PVs are also either using the `allowRemoteVolumeAcesss: "false"` policy or are otherwise bound
-      to the Node to be evacuated in their `.spec.nodeAffinity`, they are annotated with
-      `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity Controller to temporarily mark the PV as
-      accessible from any Node.
+    *   Are attached on the Node to be evacuated and marks them with `wait-for-reattach.evacuation.piraeus.io/<node-name>`
+        annotation. It will later use this annotation to wait for the PV to be reattached.
+    *   If the PV has a `linstor.csi.linbit.com/evacuation-action` annotation or the storage class has the
+        `property.linstor.csi.linbit.com/Aux/linstor.csi.linbit.com/evacuation-action` parameter set, the value is used
+        to determine the evacuation action. Possible actions are:
+        * `AffinityOverride`: Annotate the PV with `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity
+          Controller to temporarily mark the PV as accessible from any Node.
+        * `Delete`: Delete the PVC and PV, causing the **existing data** to be **lost**. This is _only_ useful for PVCs
+          storing caches or otherwise ephemeral data that get recreated automatically, for example using a StatefulSet
+          with `volumeClaimTemplates`.
+        * `None`: No special action is taken to ensure the PV can be evacuated.
+
+        If no action is specified, PVs using the `allowRemoteVolumeAccess: "false"` policy, or PVs that are otherwise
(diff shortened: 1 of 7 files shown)


docs/explanation/deletion-policies.md
```diff
@@ -68,10 +68,17 @@ In order to keep potential fail-over times short, Piraeus Datastore performs the
...
-      accessible from any Node.
+    * If the PV has a `linstor.csi.linbit.com/evacuation-action` annotation or the storage class has the
+      `property.linstor.csi.linbit.com/Aux/linstor.csi.linbit.com/evacuation-action` parameter set, the value is used
+      to determine the evacuation action. Possible actions are:
+        * `AffinityOverride`: Annotate the PV with `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity
+          Controller to temporarily mark the PV as accessible from any Node.
+        * `Delete`: Delete the PVC and PV, assuming that they get recreated on moving the workload to another node. This
+          causes the existing data to be lost.
+        * `None`: No special action is taken to ensure the PV can be evacuated.
+      If no action is specified, PVs using the `allowRemoteVolumeAccess: "false"` policy, or PVs that are otherwise
```
JoelColledge: In the GitHub preview this appears as part of the last bullet point. Perhaps we need to indent the sub-list more?

docs/explanation/deletion-policies.md
```diff
@@ -68,10 +68,17 @@ In order to keep potential fail-over times short, Piraeus Datastore performs the
...
-    * In addition, if the PVs are also either using the `allowRemoteVolumeAcesss: "false"` policy or are otherwise bound
-      to the Node to be evacuated in their `.spec.nodeAffinity`, they are annotated with
-      `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity Controller to temporarily mark the PV as
-      accessible from any Node.
+    * If the PV has a `linstor.csi.linbit.com/evacuation-action` annotation or the storage class has the
+      `property.linstor.csi.linbit.com/Aux/linstor.csi.linbit.com/evacuation-action` parameter set, the value is used
+      to determine the evacuation action. Possible actions are:
+        * `AffinityOverride`: Annotate the PV with `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity
+          Controller to temporarily mark the PV as accessible from any Node.
+        * `Delete`: Delete the PVC and PV, assuming that they get recreated on moving the workload to another node. This
```
JoelColledge: For my understanding: What component would re-create the PVC in this case? In basic usage, PVCs are manually created by applying a yaml, so in that case they won't automatically come back.

JoelColledge (commented):
Looking good :+1:

docs/explanation/deletion-policies.md
WanzenBug: Fixed. Checked with `mkdocs serve` that the output looks good.

docs/explanation/deletion-policies.md
WanzenBug: Expanded a bit on this, mainly StatefulSets and some other Operators would be the ones creating the PVCs

docs/explanation/deletion-policies.md
```diff
@@ -66,12 +66,22 @@ snapshots are removed.
...
+    *   Are attached on the Node to be evacuated and marks them with `wait-for-reattach.evacuation.piraeus.io/<node-name>`
+        annotation. It will later use this annotation to wait for the PV to be reattached.
+    *   If the PV has a `linstor.csi.linbit.com/evacuation-action` annotation or the storage class has the
+        `property.linstor.csi.linbit.com/Aux/linstor.csi.linbit.com/evacuation-action` parameter set, the value is used
+        to determine the evacuation action. Possible actions are:
+        * `AffinityOverride`: Annotate the PV with `override.piraeus.io/<nodename>`. This causes the LINSTOR Affinity
+          Controller to temporarily mark the PV as accessible from any Node.
+        * `Delete`: Delete the PVC and PV, causing the **existing data** to be **lost**. This is _only_ useful for PVCs
+          storing caches or otherwise ephemeral data that get recreated automatically, using a StatefulSets
+          `volumeClaimTemplates` or similar mechanisms.
```
JoelColledge:
```suggestion
          storing caches or otherwise ephemeral data that get recreated automatically, for example using a StatefulSet
          with `volumeClaimTemplates`.
```

Or if you want to keep it as it was, I think it should be `StatefulSet's`.

JoelColledge (dismissed):

JoelColledge (approved):

PR #871, opened 2025-09-02
rbac: split PVC and Node permissions

WanzenBug:
We do need to be able to delete PVCs for evacuation, we never need to be able to delete nodes.

Also: ensure that the helm chart is in sync regarding RBAC resources.

Also also: fix Makefile to not download yq every time


hack/copy-rbac-config-to-chart.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+set -e
+
+YQ="${YQ:-yq}"
+KUSTOMIZE="${KUSTOMIZE:-kustomize}"
+
+{
+	cat <<EOF
+# DO NOT EDIT; Automatically created by hack/copy-rbac-config-to-chart.sh
```
JoelColledge: Perhaps we should rename `hack/` to `tools/`? "Hack" sounds like a messy temporary solution...

JoelColledge (approved):

hack/copy-rbac-config-to-chart.sh
WanzenBug: Agreed. Let's see if I got everything :sweat_smile:

PR #884, opened 2025-09-26
Print additional status output with kubectl

WanzenBug:
We now show the following output

```
$ kubectl get linstorcluster/linstorcluster linstornodeconnection/example linstorsatelliteconfiguration/all-satellites linstorsatellite/k8s-10.test -owide

NAME                                       AVAILABLE   CONFIGURED   VERSION   SATELLITES   USED CAPACITY   VOLUMES   SNAPSHOTS   AGE
linstorcluster.piraeus.io/linstorcluster   True        True         1.31.2    1/1          0/11GiB         0         0           17h

NAME                                       CONFIGURED   AGE
linstornodeconnection.piraeus.io/example   True         36m

NAME                                                      SELECTOR   APPLIED   MATCHED   SATELLITES        AGE
linstorsatelliteconfiguration.piraeus.io/all-satellites              True      1         ["k8s-10.test"]   17h

NAME                                      CONNECTED   CONFIGURED   APPLIED CONFIGURATIONS   DELETION POLICY   USED CAPACITY   VOLUMES   SNAPSHOTS   STORAGE PROVIDERS                                                                        DEVICE LAYERS                                           AGE
linstorsatellite.piraeus.io/k8s-10.test   True        True         all-satellites           Retain            0/11GiB         0         0           ["DISKLESS","EBS_INIT","EBS_TARGET","FILE","FILE_THIN","LVM","LVM_THIN","REMOTE_SPDK"]   ["CACHE","DRBD","LUKS","NVME","STORAGE","WRITECACHE"]   17h
```

Also: no longer depend on `test` for the docker-build: that causes annoying long wait times for testing new deployments. We still run the tests during the pre-commit hook.


internal/controller/linstorcluster_controller.go
```diff
@@ -1010,13 +1030,87 @@ func (r *LinstorClusterReconciler) reconcileClusterState(ctx context.Context, lc
...
+	var total, free int64
+	for _, pool := range sp {
+		if pool.TotalCapacity == math.MaxInt64 {
+			// Skip all the "diskless" pools, they always report max capacity
+			continue
+		}
+
+		// LINSTOR reports KiB
+		total += pool.TotalCapacity * 1024
+		free += pool.FreeCapacity * 1024
```
JoelColledge, line 1113: It would be neat if we could use linstor's dedicated available storage API, rather than adding up all the storage pools. But that applies to resource groups, so we would have to add it to the storage class. I guess that is not possible.

internal/controller/linstorcluster_controller.go
```diff
@@ -210,6 +211,22 @@ func (r *LinstorClusterReconciler) reconcileAppliedResource(ctx context.Context,
...
+	// Extract information which config applied to which satellites
+	var currentSatellites piraeusiov1.LinstorSatelliteList
+	err = r.Client.List(ctx, &currentSatellites, &client.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	configAppliedToSatellites := make(map[string][]string)
+	for i := range currentSatellites.Items {
+		configs := strings.Split(currentSatellites.Items[i].Annotations[vars.AppliedConfigurationAnnotation], ",")
```
JoelColledge: It seems strange to me that we store and then retrieve this data to populate this field. Can't we pass the information about which configs applied to which satellites back from `kustomizeLinstorSatellite`?

api/v1/linstorsatellite_types.go
```diff
@@ -75,6 +75,34 @@ type LinstorSatelliteStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// The number of volumes in the LINSTOR Cluster.
```
JoelColledge: On this satellite/node rather than "in the cluster". Same for other fields below.

internal/controller/linstorcluster_controller.go
WanzenBug: Done. I did not do it originally, as changing the return value changes a lot of error-handling code, but that is a bad reason to pass this information through some obscure other means. :sweat_smile:

api/v1/linstorsatellite_types.go
WanzenBug: Done

internal/controller/linstorcluster_controller.go
WanzenBug: Yeah, I felt like for this status, it would be better to just aggregate the storage pools. There is no single "RG", there could be a lot of different storage classes. As this output is just purely informational, I felt this is good enough.

JoelColledge (approved):
Looks good :+1:

PR #886, opened 2025-10-02
ci: sign images

WanzenBug:
Use cosign to sign operator images using the "keyless" approach. The images are signed with a temporary identity tied to the specific github workflow that created them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   build-and-push:
     runs-on: ubuntu-22.04
+    permissions:
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@v5
@@ -21,6 +23,9 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v3
+      - name: Set up cosign
+        if: ${{ github.event_name != 'pull_request' }}
+        uses: sigstore/cosign-installer@v3
       - name: login to registry
         if: ${{ github.event_name != 'pull_request' }}
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -52,3 +58,8 @@ jobs:
           cache-to: type=gha,mode=max
           build-args: |
             VERSION=${{ github.ref_name }}-${{ github.sha }}
+      - name: Sign images
```
JoelColledge, line 61: Perhaps add a comment here with a link to an explanation of what this is doing?

JoelColledge (dismissed):
Clever stuff! Have you tried it somehow?

WanzenBug:
> Clever stuff! Have you tried it somehow?

Not in this repo directly. I initially added it [here](https://github.com/piraeusdatastore/piraeus/blob/master/.github/workflows/build-docker.yaml#L57-L61). The `jq` expression looks a bit different because we use the `docker-bake` action, but here we use `docker-build-push`.

What I did to verify that it actually runs the correct command, I went to the last [run on the `v2` branch](https://github.com/piraeusdatastore/piraeus-operator/actions/runs/18161418373/job/51692877518), expanded the `Build` step log, scrolled down to further expand the `Metadata`, which contains the json we want to consume. From that we take the digest and the image name. We ignore the tag, as we do not sign tags but actual manifests, identified by `quay.io/piraeusdatastore/piraeus-operator@sha256:....`.

WanzenBug:
Actually, thinking about it, we would not need the `xargs` stuff in this case, as there is only one image to sign...

WanzenBug:
I'll keep it as-is for now. In the future, we might push to multiple registries (not just quay.io), where we need to sign the image multiple times.

JoelColledge (approved):

PR #889, opened 2025-10-07
Configure workloads to share process namespaces

WanzenBug:
The "shareProcessNamespace" spec means that all containers in a Pod share the same process namespace. This also means that PID 1 is managed by the implicit "pause" container managed by k8s.

This container also acts as a child process reaper. This fixes issues with certain workloads, such as LINSTOR Satellites sometimes leaving behind some child processes in case of timeouts.


JoelColledge (approved):
Makes sense :+1:

PR #901, opened 2025-11-06
clusterapi: ignore more errors when fetching machine

WanzenBug:
There might be cases where the Machine annotations are present, but the Cluster API is running on another cluster. This is why we support setting a separate kube config for cluster API.

However, if a user just wants to use the Operator without the integration, we currently run into errors because theb client returns "NoKindMatched" errors. We ignore those errors now.


JoelColledge (approved):

JoelColledge:
> if a user just wants to use the Operator without the integration

Do we want to add a test for this case?

PR #903, opened 2025-11-07
Some CSI controller improvements

WanzenBug:
* Ensure that all sidecars can be patched using the environment, as that is simpler than patching the `args`
* Disable Volume Group Snapshots if the necessary API is not present, as that causes the csi-snapshotter to silently stop working.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Liveness probes for CSI sidecars
 
+### Changed
+
+- Disable group snapshots if we cannot be sure they are supported, otherwise the csi-snapshot silently breaks.
+
 ## [v2.10.0] - 2025-11-05
 
 ### Added
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Liveness probes for CSI sidecars
+
 ## [v2.10.0] - 2025-11-05
 
 ### Added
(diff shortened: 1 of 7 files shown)


docs/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Liveness probes for CSI sidecards
```
JoelColledge:
```suggestion
- Liveness probes for CSI sidecars
```

pkg/resources/cluster/patches/csi-controller-disable-volume-group-snapshots.yaml
```diff
@@ -0,0 +1,17 @@
...
+    metadata:
+      name: linstor-csi-controller
+    spec:
+      template:
+        spec:
+          containers:
+          - name: csi-snapshotter
+            env:
+            - name: FEATURE_GATES
+              value: CSIVolumeGroupSnapshot=false
```
JoelColledge: If this patch is applied then it will also remove any other FEATURE_GATES that we might later enable, right? How about having separate env vars for each? E.g. `FEATURE_GATE_CSI_VOLUME_GROUP_SNAPSHOT`?

WanzenBug: Yeah, you are right. Made it so that this is a special feature environment, and $(FEATURE_GATES) is empty by default. This means users can still override it with whatever feature gates might be added in the future.

JoelColledge (approved):

pkg/resources/cluster/patches/csi-controller-disable-volume-group-snapshots.yaml
JoelColledge: Good solution :+1:

PR #910, opened 2025-11-12
evacuation: add timeouts for re-attach step

WanzenBug:
We generally want to wait for Pods to be rescheduled and volumes to be attached before starting the evacuation. This way, LINSTOR can optimize evacuation by selecting the right node to evacuate to from the start.

In case a node has been drained before evacuation, our existing solution of only waiting for attached volumes failed: the previously drained Pods have been hanging in "Pending" state because they could not be scheduled. During evacuation, we allow rescheduling of the Pod, but since we do not wait for attachment, the LINSTOR Evacuation starts before the "Pending" Pods have been scheduled.

To fix this, we now wait for all volumes on the node to be attached. Since this could be volumes that actually have no "user" in the current cluster, this would have the cluster wait forever.

So we implement configurable timeout for this re-attach step. After the timeout elapses, we continue with evacuation without waiting for workloads to reattach.


JoelColledge (commented):
I've looked over this briefly and the code looks good.

One thing I'm puzzling about: Why do we expect a volume that was unattached before to "reattach"? How can we even call it "**re**attach" if it wasn't attached before?

WanzenBug:
> One thing I'm puzzling about: Why do we expect a volume that was unattached before to "reattach"? How can we even call it "reattach" if it wasn't attached before?

There might be situations in which a user (or automation) drains a node first. An example would be a condition like `disk-pressure` on the node. This could lead to the node being drained to clear the condition. 

If the volume has a "local-only" access policy, the evicted pods stay in "Pending", as the one node where they can start is cordoned. If we then start evacuating the node, it may look to the operator like the volumes are "unattached".

We could be more thorough by actually checking if there are Pods using the PVs, but that would require even more logic.

You are right that "reattach" is perhaps wrongly worded. I am open to suggestions.

api/v1/evacuation_strategy.go
```diff
@@ -0,0 +1,19 @@
...
+	// AttachedVolumeReattachTimeout configures how long evacuation waits for attached volumes to reattach on
+	// different nodes. Setting this to 0 disable this evacuation step.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:="5m"
+	AttachedVolumeReattachTimeout metav1.Duration `json:"attachedVolumeReattachTimeout"`
+	// UnattachedVolumeReattachTimeout configures how long evacuation waits for unattached volumes to attach on
+	// different nodes. Setting this to 0 disable this evacuation step.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:="5m"
+	UnattachedVolumeReattachTimeout metav1.Duration `json:"unattachedVolumeReattachTimeout"`
```
JoelColledge: > There might be situations in which a user (or automation) drains a node first...

Thanks for the explanation. So we're assuming that user's just don't have PVs without a corresponding pod, and if they do, at least evacuation will work after the timeout.

> You are right that "reattach" is perhaps wrongly worded. I am open to suggestions.

Use `Attach` instead of `Reattach` in that one case?

```
AttachedVolumeReattachTimeout
UnattachedVolumeAttachTimeout
```

Or maybe this is sufficient? Would that be misunderstood as referring to the time to detach?

```
AttachedVolumeTimeout
UnattachedVolumeTimeout
```

pkg/evacuation/evacuation.go
```diff
@@ -292,26 +312,54 @@ func waitForReattach(ctx context.Context, cl client.Client, nodeName string) (st
...
+			waitedForReattachSince, err = time.Parse(time.RFC3339, val)
+			if err != nil {
+				return "", false, err
+			}
+		}
+
+		switch pv.Annotations[vars.PersistentVolumeWaitForReattachAnnotationPrefix+"/"+nodeName] {
+		case "true":
+			if now.Sub(waitedForReattachSince) >= evacuationStrategy.AttachedVolumeReattachTimeout.Duration {
+				continue
```
JoelColledge, line 345: When we carry out the evacuation with PVs which did not re-attach within the timeout, it would be good to log the affected PVs.

pkg/evacuation/evacuation.go
```diff
@@ -292,26 +312,54 @@ func waitForReattach(ctx context.Context, cl client.Client, nodeName string) (st
...
 		}
 	}
 
 	if len(unattachedPVs) > 0 {
-		return fmt.Sprintf("Waiting for PVs to reattach on other nodes: %s", strings.Join(unattachedPVs, ", ")), false, nil
+		unattachedPVNames := make([]string, 0, len(unattachedPVs))
+		for _, pv := range unattachedPVs {
+			if _, ok := pv.Annotations[vars.PersistentVolumeWaitForReattachSinceAnnotationPrefix+"/"+nodeName]; !ok {
+				// NB: pv.Annotations cannot be null, as only PVs with the "wait-for-reattach" annotation are considered
+				pv.Annotations[vars.PersistentVolumeWaitForReattachSinceAnnotationPrefix+"/"+nodeName] = now.Format(time.RFC3339)
```
JoelColledge, line 379: Why do we annotate every PV with the same value? Isn't one timestamp on the node simpler, indicating when we started waiting for re-attach?

WanzenBug: I initially started with that. I then switched to the annotations on the PV because it makes later reconciliation simpler.

We could run into a scenario, where we timed out in this step, we continue the evacuation, and then the node is already removed from k8s, but we still have the satellite resources. We "lost" the information that we have already waited long enough and would need to do some extra checks that the node to evacuate even exists.

In the end, saving that information directly on the PV was the lazy option :smile:

api/v1/evacuation_strategy.go
WanzenBug: > Use Attach instead of Reattach in that one case?

Applied that. 

> ```
> AttachedVolumeTimeout
> UnattachedVolumeTimeout
> ```

For me, this sounds too much like this is the overall timeout for evacuation, i.e. if the volume does not "evacuate" within 5min it is somehow force-removed.

pkg/evacuation/evacuation.go
WanzenBug: Will do that. I'll add some Kubernetes "Events", which are displayed to users on running `kubectl describe pv ...` and similar.

docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -287,6 +287,32 @@ spec:
  deletionPolicy: Evacuate
 ```
 
+### `spec.evacuationStrategy`
+
+Configure timeouts for Satellite Evacuation. To have any effect, `deletionPolicy: Delete` is required.
+
+* `attachedVolumeReattachTimeout`: sets the duration the evacuation waits for volumes that were attached before draining
+  the Node to be attached on a new node. Defaults to 5 minutes. Can be set to 0 to disable waiting.
+* `unattachedVolumeReattachTimeout`: sets the duration the evacuation waits for volumes that were not attached before
```
JoelColledge:
```suggestion
* `unattachedVolumeAttachTimeout`: sets the duration the evacuation waits for volumes that were not attached before
```

docs/reference/linstorsatelliteconfiguration.md
```diff
@@ -287,6 +287,32 @@ spec:
...
+```yaml
+apiVersion: piraeus.io/v1
+kind: LinstorSatelliteConfiguration
+metadata:
+  name: evacuation-timeout
+spec:
+  deletionPolicy: Evacuate
+  evacuationStrategy:
+    attachedVolumeReattachTimeout: "1h"
+    unattachedVolumeReattachTimeout: "30m"
```
JoelColledge:
```suggestion
    unattachedVolumeAttachTimeout: "30m"
```

pkg/evacuation/evacuation.go
```diff
@@ -477,3 +580,12 @@ func isAttachableOnOtherNode(pv *corev1.PersistentVolume, currentNodeName string
 
 	return false
 }
+
+// PVCorPVEventf records an event, either for the PVC or for the PV, if no PVC is currently bound.
+func PVCorPVEventf(recorder record.EventRecorder, pv *corev1.PersistentVolume, eventtype, reason, message string, args ...interface{}) {
```
JoelColledge: Shouldn't it strictly be so?

```suggestion
// PVCOrPVEventf records an event, either for the PVC or for the PV, if no PVC is currently bound.
func PVCOrPVEventf(recorder record.EventRecorder, pv *corev1.PersistentVolume, eventtype, reason, message string, args ...interface{}) {
```

Aside: I prefer `PvcOrPvEventf` or even better `pvc_or_pv_eventf`, but we're not going to change the Go world now.

JoelColledge (commented):
Looking good :+1:

JoelColledge (approved):

PR #941, opened 2026-01-15
Add namespace values

cometship:
This pull request adds missing `namespace: {{ .Release.Namespace }}`. Important for Helm template in Terraform/ArgoCD. Else they get deployed in the `default` namespace.

See: https://github.com/helm/helm/issues/3553#issuecomment-2235841549


WanzenBug (approved):
Thanks!

PR #959, opened 2026-02-24
Let satellites daemonset ignore NoSchedule taints

WanzenBug:
If a user added a NoSchedule taint to a Node (for example for planned removal of a LINSTOR Satellite), and then accidentally deleted the Satellite Pod before the Operator completed the Satellites' removal, we got stuck in a situation where:

* There is a LinstorSatellite resource
* There is no Satellite Pod
* The operator tries to reconcile, but fails because the Pod cannot get rescheduled.

This is caused by LinstorSatellite resource effectively working like they directly control the specific Pod, but that Pod is managed by the indirection of the DaemonSet.

To resolve this, always let the Satellite DaemonSet tolerate all NoSchedule taints. This way, the Pod gets restarted until the LinstorSatellite is actually gone.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Updated NFS server template to work with NFS Ganesha instances with monitoring enabled.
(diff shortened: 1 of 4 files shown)


JoelColledge (approved):
Looks good 👍 

Took a bit longer to review because I actually built and tested it this time. Good to get familiar with those processes again.

PR #960, opened 2026-02-25
ci: fix image signing step when commit message contains apostrophe

JoelColledge:
A commit message containing an isolated `'` character, that is a single quote or apostrophe, caused the "Sign images" step to fail. The error message was extremely cryptic:

```
Run jq '."containerimage.digest" as $DIGEST | ."image.name" | split(":")[0] | "\(.)@\($DIGEST)"' -r <<<'{
...
/home/runner/work/_temp/4a1f55e3-cea3-4bb3-b356-51419ed4b652.sh: line 124: syntax error near unexpected token `('
```

The underlying issue was that the `'` finished the string. The shell parser then failed when it came to the next `(`.

Note that substrings like `'5'` did not cause problems because the shell concatenates the strings.

Fix this by evaluating the expression into an environment variable first, as shown at
https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,9 @@ jobs:
         # Take the digest we just pushed to the registry and sign the image using "keyless" signing:
         #   https://docs.sigstore.dev/cosign/signing/overview/
         # With this, users can verify that the image was actually created by this github workflow.
+        env:
+          BUILD_METADATA: ${{ steps.build.outputs.metadata }}
         run: |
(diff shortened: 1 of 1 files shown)


JoelColledge:
Fixes this failure: https://github.com/piraeusdatastore/piraeus-operator/actions/runs/22394265570/job/64837400851

I have only tested the new script locally. I have not tested this in the context of GitHub Actions.

WanzenBug (approved):
Thanks! Kind of unbelievable that this is the best way to pass things into an action step...

PR #990, opened 2026-05-10
fix: handle dirty Kubernetes minor versions

immanuwell:
Tiny fix for Kubernetes version discovery.

Kubernetes minor versions can be numeric with a trailing `+`; client-go documents this as `numeric possibly followed by "+"`. Right now `ServerVersion()` uses `strconv.Atoi` directly, so `34+` falls back to `v1.20`. That can make version-gated behavior pick the wrong path, oof.

Repro:
- serve `/version` with `major: "1"`, `minor: "34+"`
- call `APIDiscoveryClient.ServerVersion()`
- before: returns the fallback `v1.20`
- after: returns `v1.34`

Checks:
- `go test ./pkg/utils`
- `make test`

I searched related issues/PRs for Kubernetes version fallback / minor parsing and did not find a direct match.

### DIFF
--- a/pkg/utils/version.go
+++ b/pkg/utils/version.go
@@ -4,6 +4,7 @@ import (
 	"cmp"
 	"fmt"
 	"strconv"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


WanzenBug (approved):
Thanks! LGTM!

Issue #393, opened 2023-01-19
`controllerEndpoint` issue, only first endpoint is used


jonathon2nd:
We have the piraeus-operator installed on k8s, with linstor cluster on host.
Deployed with helm.
REPO URL: https://github.com/piraeusdatastore/piraeus-operator.git
TARGET REVISION: v1.10.1

Values
```
---
operator:
  controller:
    enabled: false
  satelliteSet:
    kernelModuleInjectionImage: quay.io/piraeusdatastore/drbd9-centos8:v9.2.1
    satelliteImage: quay.io/piraeusdatastore/piraeus-server:v1.20.0-1
etcd: 
  enabled: false
controllerEndpoint: http://10.2.4.10:3370/,http://10.2.4.12:3370/,http://10.2.4.11:3370/
```

So you can see we have `controllerEndpoint: http://10.2.4.10:3370/,http://10.2.4.12:3370/,http://10.2.4.11:3370/`, the problem is it only works if the first host in the list is the controller. When we did upgrades on the cluster and the controller failed over to another host, the  operator failed, as it no longer had connection. It did not try the other two hosts in the list.

I don't have error logs now, as this happened a little bit again, and the errors were very basic and I did not record.

jonathon2nd:
Still happening with V2.
![image](https://github.com/piraeusdatastore/piraeus-operator/assets/52681917/6f30e438-0978-48a7-bfeb-2f8893e475af)

```
apiVersion: piraeus.io/v1
kind: LinstorCluster
metadata:
  name: linstorcluster
spec:
  externalController:
    url: >-
      http://10.2.4.10:3370/,http://10.2.4.12:3370/,http://10.2.4.11:3370/,http://10.2.4.13:3370/
  nodeAffinity:
    nodeSelectorTerms:
      - matchExpressions:
          - key: node-role.kubernetes.io/control-plane
            operator: DoesNotExist
status:
  conditions:
    - lastTransitionTime: '2024-06-13T21:27:25Z'
      message: Resources applied
      observedGeneration: 5
      reason: AsExpected
      status: 'True'
      type: Applied
    - lastTransitionTime: '2024-06-13T21:05:54Z'
      message: >-
        Error: Get "http://10.2.4.10:3370/v1/controller/version": dial tcp
        10.2.4.10:3370: connect: connection refused
      observedGeneration: 5
      reason: Error
      status: 'False'
      type: Available
[...]

PR #999, opened 2026-05-21
fix: accept external controller failover URLs

immanuwell:
Tiny fix for a real external-controller gotcha.

The webhook parsed `spec.externalController.url` as one URL. But the operator later splits it on `,` and passes the entries to the LINSTOR client. So this valid failover list got rejected at admission, kinda awkward:

```yaml
spec:
  externalController:
    url: http://linstor-controller.example.com:3370,linstor://backup-controller.example.com:3370
```

Repro before this patch:
`kubectl apply -f` that LinstorCluster, webhook rejects it with `invalid port ":3370,linstor:" after host`.

Fix: validate with the same LINSTOR client parser used at runtime.

Tests:
`KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.36.0-linux-amd64" go test ./...`

Related to #393.

### DIFF
--- a/internal/webhook/v1/linstorcluster_webhook.go
+++ b/internal/webhook/v1/linstorcluster_webhook.go
@@ -18,9 +18,10 @@ package v1
 
 import (
 	"context"
-	"net/url"
 	"strconv"
+	"strings"
 
+	lapi "github.com/LINBIT/golinstor/client"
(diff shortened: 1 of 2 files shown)


WanzenBug (approved):

WanzenBug:
Thanks!

PR #1001, opened 2026-05-24
fix: render Helm image pull secrets

immanuwell:
Small Helm chart papercut.

`imagePullSecrets` is in `values.yaml`, but setting it was a no-op. So private registry users could pass a pull secret and the operator Deployment still would not use it. Kinda rough.

Repro:
```sh
helm template test charts/piraeus \
  --namespace piraeus-datastore \
  --set imagePullSecrets[0].name=registry-creds | rg "imagePullSecrets|registry-creds"
```

Before: no output.
After: the Deployment pod spec includes:
```yaml
imagePullSecrets:
  - name: registry-creds
```

Checked:
```sh
helm lint charts/piraeus
helm template test charts/piraeus --namespace piraeus-datastore
```

No direct related issue found, afaict.

### DIFF
--- a/charts/piraeus/templates/deployment.yaml
+++ b/charts/piraeus/templates/deployment.yaml
@@ -107,6 +107,10 @@ spec:
         {{ toYaml . | nindent 8 }}
         {{- end }}
       serviceAccountName: {{ include "piraeus-operator.serviceAccountName" . }}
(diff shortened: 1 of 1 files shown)


WanzenBug (approved):

WanzenBug:
Thanks!

Issue #614, opened 2024-03-02
TLS for DRBD Replication doesn't work


retinio:
Hi! 
I try to configure tls for DRBD by this [manual](https://github.com/piraeusdatastore/piraeus-operator/blob/v2/docs/how-to/drbd-tls.md)
TLS for internal traffic is enable:
```
kind: LinstorSatelliteConfiguration
spec:
  internalTLS:
     tlsHandshakeDaemon: true
     secretName: linstor-satellite-internal-tls
```

λ **kubectl exec -n linstor deploy/linstor-controller -- linstor node list**
+---------------------------------------------------------------+
| Node      | NodeType  | Addresses                            | State  |
|======================================|
| worker-01 | SATELLITE | 192.168.160.20:3367 (SSL) | Online |
| worker-02 | SATELLITE | 192.168.160.21:3367 (SSL) | Online |
| worker-03 | SATELLITE | 192.168.160.22:3367 (SSL) | Online |
+---------------------------------------------------------------+

But drdb doesn't connect  to each other
λ **kubectl exec -n linstor deploy/linstor-controller -- linstor r l**
+-------------------------------------------------------------------------------------------------------------------------------------------+
| ResourceName                                                 | Node         | Port  | Usage    | Conns                                            |      State |            |
|===================================================================================
| pvc-4973e04e-44cf-49fe-9094-98dfbfda10d5 | worker-01 | 7000 | Unused | StandAlone(worker-03,worker-02) |   UpToDate        | 
| pvc-4973e04e-44cf-49fe-9094-98dfbfda10d5 | worker-02 | 7000 | Unused | StandAlone(worker-03,worker-01) | TieBreaker         | 
| pvc-4973e04e-44cf-49fe-9094-98dfbfda10d5 | worker-03 | 7000 | InUse    | StandAlone(worker-01,worker-02) |   UpToDate        |
+-------------------------------------------------------------------------------------------------------------------------------------------+
 ktls-utils containers have errors:
λ **kubectl -n linstor logs -l app.kubernetes.io/component=linstor-satellite -c ktls-utils**
```
tlshd[29]: gnutls: The TLS connection was non-properly terminated. (-110)
tlshd[29]: Handshake with 'worker-02' (192.168.160.21) failed
tlshd[34]: gnutls: Error in the certificate. (-43)
tlshd[34]: Handshake with 'worker-03' (192.168.160.22) failed
tlshd[32]: gnutls: Error in the certificate. (-43)
[...]

PR #1007, opened 2026-05-30
docs: fix typo and header inconsistencies in TLS docs

immanuwell:
Two small doc bugs in the TLS guides.

Typo in `docs/how-to/drbd-tls.md`:

```
You should see an equal, nonzero number of `TlsRxSw` and `TlsRxSw`.
```

First counter should be `TlsTxSw`. The code block right below already shows both `TlsTxSw` and `TlsRxSw` as separate counters with value 4, so the text just contradicts itself.

Missing leading dot in two section headers in `docs/reference/linstorsatelliteconfiguration.md` - `spec.deletionPolicy` and `spec.evacuationStrategy` vs every other header in the file which uses `.spec.*` prefix.

Related: #614


WanzenBug (approved):
LGTM, thanks!

PR #1009, opened 2026-06-06
fix: tolerate custom NoSchedule taints on node DaemonSets

### LINKED ISSUE #957: Mismatch between LinstorSatellite tolerations and actual Pods
When a node is evacuated, a user might use something like `kubectl taint node <node> piraeus.io/remove-satellite:NoSchedule` and then delete the `LinstorSatellite` resource. This means that after evacuation, the Satellite is not recreated. 

However, there is the chance that critical pods (linstor-csi-node, linstor-satellite, etc..) get removed first. This then blocks evacuation, as the DaemonSets get recreated, but because of the above taint, the Pod cannot be scheduled. So we have a satellite that should be evacuated, but cannot be because the Pod was removed and it does not get recreated because it can't be scheduled.

We should consider updating the Pod tolerations so that in this situations, critical pods can get recreated

immanuwell:
Fixes #957

A `LinstorSatellite` may stay on a node with a custom `NoSchedule` taint, but the node DaemonSets do not tolerate that taint. If `linstor-csi-node`, `ha-controller`, or `linstor-csi-nfs-server` gets recreated there, the replacement pod just sits Pending. Bit rough.

This adds the same broad `NoSchedule` toleration to those DaemonSets.

Repro:
1. Create a `LinstorCluster` and wait until satellites are up.
2. Taint one node:
   `kubectl taint node <node> piraeus.io/remove-satellite:NoSchedule`
3. Delete one pod on that node, for example:
   `kubectl delete pod -n piraeus-datastore <linstor-csi-node-pod>`
4. Before this patch the replacement stays Pending, after this patch it comes back.

Tests:
`go test ./internal/controller -run '^(TestPatches|TestNodeDaemonSetsTolerateNoSchedule)$'`

### DIFF
--- a/internal/controller/linstorcluster_controller.go
+++ b/internal/controller/linstorcluster_controller.go
@@ -621,7 +621,7 @@ func (r *LinstorClusterReconciler) kustomizeCSINodeResources(lcluster *piraeusio
 		patches = append(patches, p...)
 	}
 
-	t := tolerations.MergeTolerations(tolerations.HAControllerTolerations, lcluster.Spec.Tolerations)
+	t := tolerations.MergeTolerations(tolerations.HAControllerTolerations, tolerations.NoScheduleToleration, lcluster.Spec.Tolerations)
 	p, err := TolerationsPatch("DaemonSet", "linstor-csi-node", t)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


WanzenBug (changes requested):
I'm not sure this is actually what we need. At the very least, we should actually bind scheduling to the existence of satellite pods.

Right now, we would ignore all NoSchedule taints. But that is not actually what happens with the Satellites. We do it there, because we already have the `LinstorSatellite` resource that handles the scheduling.

So if we add those tolerations, we also need a inter-pod affinity for `app.kubernetes.io/component: linstor-satellite`

immanuwell:
@WanzenBug got it, what do you think about 14ef64a3 ?

WanzenBug:
> @WanzenBug got it, what do you think about https://github.com/piraeusdatastore/piraeus-operator/commit/14ef64a3f6af49a33dc2f8e4ce6a49a77ef55a88 ?

I just gave it a test, but I still get pods hanging in Pending after I remove the satellite. It looks like DaemonSets ignore the Pod Affinity :/

So if we proceed with this change we get Pods hanging in Pending on all Unschedulable nodes that have no Satellite, and if we do not apply this change we get Pods in Pending on when the Node is using some custom :NoSchedule taint :cry:

immanuwell:
@WanzenBug thanks, I should have thought about it more carefully
so I am going to close this PR

PR #1010, opened 2026-06-08
Add checks before registring storage pools

WanzenBug:
* Check if the LVM pool or zpool exists before registering a storage pool.
* If not, and a device source is set, check that all devices actually exist.


JoelColledge (dismissed):
Looks like a great usability improvement.

WanzenBug:
@JoelColledge rebased and fixed the conflict. Please re-review :pray:

JoelColledge (approved):

PR #1017, opened 2026-06-25
hook klog into controller-runtime zap logger

camh-:
Use `klog.SetLogger()` to have klog send its logs to the zap logger
configured for the controller runtime. Without this, the components that
use klog, such as leader-election, will log in the default klog format
even if the controller has been configured to use a different format as
can be done with `--zap-encoder=json`.


WanzenBug (approved):

WanzenBug:
LGTM. Thanks!

PR #1018, opened 2026-06-27
make csi-node container cli values patchable

camh-:
Make the values of the flags passed to containers of the csi-node
DaemonSet patchable by taking those values from environment variables.
Environment variables are much easier to patch with LinstorCluster
patches than the args array. This is the same way it is done with the
csi-controller containers.

Update docs/how-to/k0s.md to patch the `KUBELET_REGISTRATION_PATH`
environment variable instead of the whole set of args.

### DIFF
--- a/docs/how-to/k0s.md
+++ b/docs/how-to/k0s.md
@@ -29,11 +29,9 @@ spec:
               name: publish-dir
               mountPropagation: Bidirectional
           - name: csi-node-driver-registrar
-            args:
-            - --v=5
-            - --csi-address=/csi/csi.sock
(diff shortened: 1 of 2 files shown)


camh-:
I've created this change as I have found some of the csi-node logs to be a bit too chatty for my liking and wanted to turn them down a bit, but it is hard to patch an element of an array (the args with --v=5). The controller has a method of patching the env var value, so figured it would be good to have on the node too.

WanzenBug (approved):

WanzenBug:
Thanks!

PR #1019, opened 2026-06-29
limit concurrent satellite evacuations

WanzenBug:
Add LinstorCluster.spec.maxConcurrentEvacuations to cap how many Satellites evacuate at the same time when using the Evacuate deletion policy. A value of 0 (the default) keeps the previous unlimited behavior.

Admission runs as the first step of EvacuateSatellite, which now owns the SatelliteEvacuated condition. Candidates are derived from that condition across all Satellites, so the same gate applies to both direct LinstorSatellite removals and ClusterAPI-driven node removals. Candidates are ordered in-progress first, then oldest waiter, then by name, and the first N are admitted; the rest report a Waiting condition without starting any evacuation step. For ClusterAPI nodes this keeps the pre-drain hook in place, so a waiting node stays running until a slot frees.

A missing LinstorCluster (external controller or teardown) is treated as no limit, so evacuation is never blocked.

### DIFF
--- a/pkg/evacuation/evacuation.go
+++ b/pkg/evacuation/evacuation.go
@@ -40,18 +40,29 @@ const (
 // SatelliteEvacuatedCondition tracks the progress of evacuating volumes off a Satellite.
 const SatelliteEvacuatedCondition = "SatelliteEvacuated"
 
-// admissionMu serializes evacuation admission decisions so concurrent reconciles compute a consistent view
-// of how many Satellites are currently evacuating.
-var admissionMu sync.Mutex
+// Evacuator moves LINSTOR resources off a Satellite that is being removed. It bundles the Kubernetes and
+// ClusterAPI clients, which are stable for the lifetime of the reconciler; the per-cluster LINSTOR client is
+// passed to EvacuateSatellite instead.
+type Evacuator struct {
+	// Client reads and updates PVs, VolumeAttachments and LinstorSatellites.
+	Client client.Client
+	// Recorder records events on the affected PVCs/PVs.
+	Recorder record.EventRecorder
+	// MachineClient releases the ClusterAPI Machine drain/terminate hooks as evacuation progresses.
+	MachineClient *clusterapi.Client
+
+	// admissionMu serializes evacuation admission decisions so concurrent reconciles compute a consistent
+	// view of how many Satellites are currently evacuating.
+	admissionMu sync.Mutex
+}
 
 // EvacuateSatellite ensures all LINSTOR Resources have been moved to other nodes, honoring the cluster-wide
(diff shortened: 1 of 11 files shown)


pkg/evacuation/evacuation.go
```diff
@@ -36,11 +37,133 @@ const (
...
+// of how many Satellites are currently evacuating.
+var admissionMu sync.Mutex
+
+// EvacuateSatellite ensures all LINSTOR Resources have been moved to other nodes, honoring the cluster-wide
+// evacuation concurrency limit, and records the progress on the SatelliteEvacuatedCondition.
+//
+// Returns a message indicating the current progress and a bool indicating if evacuation is complete. The
+// function should be called repeatedly until completion of the evacuation.
+func EvacuateSatellite(ctx context.Context, cl client.Client, lclient *lapi.Client, recorder record.EventRecorder, node *lapi.Node, machineClient *clusterapi.Client, machine *clusterapi.Machine, evacuationStrategy *piraeusiov1.EvacuationStrategy, conds conditions.Conditions, maxConcurrent int) (bool, error) {
+	// Step 0: respect the cluster-wide evacuation concurrency limit.
```
JoelColledge: "Step 0" only? Not obvious that this is part of the sequence that continues in `evacuateSatellite`.

pkg/evacuation/evacuation.go
```diff
@@ -36,11 +37,133 @@ const (
...
+// admissionMu serializes evacuation admission decisions so concurrent reconciles compute a consistent view
+// of how many Satellites are currently evacuating.
+var admissionMu sync.Mutex
+
+// EvacuateSatellite ensures all LINSTOR Resources have been moved to other nodes, honoring the cluster-wide
+// evacuation concurrency limit, and records the progress on the SatelliteEvacuatedCondition.
+//
+// Returns a message indicating the current progress and a bool indicating if evacuation is complete. The
+// function should be called repeatedly until completion of the evacuation.
+func EvacuateSatellite(ctx context.Context, cl client.Client, lclient *lapi.Client, recorder record.EventRecorder, node *lapi.Node, machineClient *clusterapi.Client, machine *clusterapi.Machine, evacuationStrategy *piraeusiov1.EvacuationStrategy, conds conditions.Conditions, maxConcurrent int) (bool, error) {
```
JoelColledge: This list of arguments is very long. Any ideas how to reduce it? I don't have any ideas.

JoelColledge (commented):
Looks good overall 👍

pkg/evacuation/evacuation.go
WanzenBug: Moved the common things into an "evacuator" struct. I feel like this makes it a bit cleaner.

JoelColledge (approved):

PR #1044, opened 2026-08-12
Support shared LVM storage pools

WanzenBug:
Allow marking a LVM storage pool as backed by storage shared between nodes, such as a SAN. Setting the new "lvmPool.sharedSpace" field registers the pool with the given shared space name in LINSTOR. Shared VGs must be set up manually by the user, they cannot be created from source devices. LINSTOR only supports shared pools for plain LVM, not LVM thin.

Optionally, "lvmPool.externalLocking" skips LINSTOR's internal locking in favor of an external lock manager such as lvmlockd. In that case the satellite pod is patched to mount the host's /run/lvmlockd.pid; the lvmlockd sockets in /run/lvm are already part of the base satellite mounts.

The webhook validates the shared space name against LINSTOR's rules, rejects sources on shared pools, requires a shared space for external locking, and forbids changing either field on an existing pool.

### DIFF
--- a/internal/controller/patches.go
+++ b/internal/controller/patches.go
@@ -313,6 +313,18 @@ func SatellitePrecompiledModulePatch() ([]kusttypes.Patch, error) {
 	)
 }
 
+// SatelliteLvmlockdPatch mounts the host's lvmlockd pid file into the satellite container.
+// The lvmlockd sockets are covered by the /run/lvm mount in the base resources, which LVM
+// always requires. The pid file lives directly in /run; it is patched in separately so we
+// don't have to expose the host's entire /run tmpfs just for this one optional file.
+func SatelliteLvmlockdPatch() ([]kusttypes.Patch, error) {
+	return render(
+		satellite.Resources,
+		"patches/lvmlockd.yaml",
+		nil,
+	)
+}
+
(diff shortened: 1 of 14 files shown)


internal/controller/patches.go
```diff
@@ -313,6 +313,16 @@ func SatellitePrecompiledModulePatch() ([]kusttypes.Patch, error) {
 	)
 }
 
+// SatelliteLvmlockdPatch mounts the host's lvmlockd pid file into the satellite container.
+// The lvmlockd sockets in /run/lvm are already part of the base satellite resources.
+func SatelliteLvmlockdPatch() ([]kusttypes.Patch, error) {
```
JoelColledge, line 320: Worth adding a brief justification for why the sockets are part of the base resource, but the pid file is patched dynamically?

JoelColledge (dismissed):
+332 **-0** 😯

internal/controller/patches.go
WanzenBug: Amended. TLDR: /run/lvm is "LVM" only, but I can't pull in /run/lvmlockd.pid if it does not exist, and pulling in all of /run seems excessive and would perhaps expose some "secrets" we rather not get easy access to.

JoelColledge (approved):

PR #1045, opened 2026-08-14
Raise minimum supported Kubernetes to v1.30

WanzenBug:
The oldest Kubernetes version still under vendor support (OpenShift 4.18 EUS, Rancher 2.12, EKS/GKE/AKS extended support) is v1.31, so a v1.30 floor covers every commercially supported distribution.

This allows removing all version-gated compatibility code:

* Drop the CSIDriver patch that removed spec.seLinuxMount on clusters older than v1.25, along with the server version detection it needed.
* Run the compat tests with the regular setup-envtest: binaries for v1.30 are available from the GitHub release channel, so the extra setup-envtest pinned to release-0.17 and the deprecated GCS bucket are no longer needed.


JoelColledge (approved):
Good cleanup 👍

PR #1053, opened 2026-08-27
Remove cluster-wide Secret RBAC, document S3 backups

WanzenBug:
The CSI controller ServiceAccount had cluster-wide permission to read all Secrets, only used by the csi-snapshotter sidecar to fetch Secrets referenced in a VolumeSnapshotClass. Drop the rule and require users to grant access to the specific Secret instead, matching the upstream external-snapshotter default.

With no deployed ClusterRole granting Secret access anymore, the Operator itself no longer needs cluster-wide Secret permissions either: move them to a namespaced Role, as the Operator only manages Secrets in its own namespace.

Add a how-to guide for S3 backups, covering the new explicit grant, and upgrade notes for existing users of snapshot secrets.

### DIFF
--- a/charts/piraeus/templates/rbac.yaml
+++ b/charts/piraeus/templates/rbac.yaml
@@ -15,7 +15,6 @@ rules:
     - events
     - persistentvolumes
     - pods
-    - secrets
     - serviceaccounts
     - services
   verbs:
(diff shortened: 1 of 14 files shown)


JoelColledge (approved):
All looks good to me.

PR #1055, opened 2026-08-31
Certificate rotation for LINSTOR components

### LINKED ISSUE #701: Controller unreachable after cert-manager certificates expire
**Environment details**
- Kubernetes v1.29.4
- pireaeus-operator 2.5.1
- cert-manager v1.14.5

**Description**
I've setup both the internal and api TLS using certmanager as described in how-to docs.  It seems that the certificates are valid for 3 month upon issue.  Certmanager renews the certificates automatically 1 month before expiry, but the deployed linstor pods still use the old certificates.  When the original certificates expire, cluster communications are broken.  I had to manually delete the controller, satellite and csi pods so that they spin up again using the renewed certs.

**Desired behavior**
Controller, satellite and csi pods should reload certs or restart after renewal.

WanzenBug:
Adds liveness probes for LINSTOR Controller and Satellites, that ensure the containers get restarted once Certificates rotate.

To this end, we compare the k8s managed `*.pem/*.crt` files with what gets imported into the java keystore.

### DIFF
--- a/docs/how-to/internal-tls.md
+++ b/docs/how-to/internal-tls.md
@@ -174,6 +174,15 @@ k8s_secret_trusted_by satellite-tls controller-tls
 # satellite-tls.tls.crt: OK
 ```
 
+## Certificate Rotation
+
+LINSTOR only reads certificates during start-up. To ensure rotated certificates are picked up, for example after
+cert-manager renewed a certificate, the Operator deploys the Satellite with a liveness probe that detects
+stale certificates. When the probe fails, the container is restarted, loading the new certificates.
+
+The restart does not interrupt replication: DRBD® connections are handled entirely in the kernel and remain up while
+the Satellite container restarts.
+
 All available options are documented in the reference for
 [`LinstorCluster`](../reference/linstorcluster.md#specinternaltls) and
 [`LinstorSatelliteConfiguration`](../reference/linstorsatelliteconfiguration.md#specinternaltls).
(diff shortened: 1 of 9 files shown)


WanzenBug:
@kvaps I know you are using Reloader in CozyStack to achieve this currently. What do you think? All the `golinstor` based components already have per-connection loading of the certificates, so those work fine with rotation, but LINSTOR needs this restart.

docs/how-to/internal-tls.md
```diff
@@ -174,6 +174,15 @@ k8s_secret_trusted_by satellite-tls controller-tls
...
 ```
 
+## Certificate Rotation
+
+LINSTOR only reads certificates during start-up. To ensure rotated certificates are picked up, for example after
+cert-manager renewed a certificate, the Operator deploys the Satellite with a liveness probe that detects
+stale certificates. When the probe fails, the container is restarted, loading the new certificates.
+
+The restart does not interrupt replication: DRBD® connections are handled entirely in the kernel and remain up while
+the Satellite container restarts.
```
JoelColledge, line 184: Does this work when host networking is not used? I would have expected the restart to involve the network namespace being torn down, cause a temporary loss of connection for DRBD.

WanzenBug: The net namespace survives because just the container is restarted. For example the drbd-reactor container keeps running, which is part of the same Pod.

JoelColledge: 💡

JoelColledge (approved):

PR #1065, opened 2026-09-22
Fix list rendering in docs

WanzenBug:
We do not use the CommonMark dialect for our docs pages, so it is easy to have something that looks correct, even when rendered on github, that then looks wrong on our docs pages.

Fix this, at least for lists, by comparing the rendered list items between the two markdown formats.

Also fix the docs affected by this. Example of "wrong" rendering here: https://piraeus.io/docs/v2.12.0/reference/linstorsatelliteconfiguration/#specproperties


JoelColledge (approved):
Impressive that this check works with so little code.

